PR #4, opened 2016-10-04
Add some basic structure to the README

pipermerriam:
This adds some basic structure to the README.  I'm thinking that while issues are an *ok* way to discuss stuff we might do well by opening pull requests into the repository itself and having the discussion via feedback on the PR.  this way we can have threaded discussions on specific topics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,26 @@
 # epm
-Ethereum Package Manager
+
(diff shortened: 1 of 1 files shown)


tcoulter:
Better than what's there. Merged.

Issue #2, opened 2016-10-01
Package manifest discussion


tcoulter:
Before deciding on how package management may work, we should first decide on what information should be stored in the package manifest (think, `package.json`). Afterward, we should decide which format this data should be stored in.
## Possible data to be stored within the manifest
- package name
- author
- version (preferably [semver](http://semver.org/))
- description
- contract metadata
  - abi
  - unlinked binary
  - deployed address if applicable
  - name/address pairs for linked libraries
  - compiler version for bytecode verification
  - key/value pairs of the sha3 hash of events this contract can emit with the abi definition of that event (i.e., the contract metadata should contain information for all events that can be triggered by this contract, including those triggered by linked libraries)
- external project URIs (homepage, repository uri, etc.)
- dependencies (i.e., references to other packages and versions)
- updated time (likely added by the packager)
## Manifest format

As far as package formats, I've made some assumptions:
1. Package manifests are primarily for non-contract actors and will not be consumed by contracts themselves.
2. Following from that, we will likely store the manifest on some distributed protocol like IPFS or Swarm (although this is jumping ahead in the discussion).
3. Since our off chain actors are primarily software, the message format needs to be easily readable by most programming languages.
4. That said, it's likely human actors will write the manifest, meaning it needs to be in a format humans can easily understand and comprehend.
5. Contrasting to the above, we need a format that takes up the least amount of space possible so that it costs the least to make available

All that said, my preferred format is JSON. Something like [MessagePack](http://msgpack.org/index.html) could be interesting however; it's less consumable by humans but [can easily be converted to and from JSON](https://github.com/ludocode/msgpack-tools)

j-h-scheufen:
I would vote for JSON as the preferred format as well.
We have been maintaining a manifest for Eris smart contract packages (we call them contract bundles) that is a slightly modified version of a NPM package.json. The metadata is used for documentation purposes only at the moment as no dependency manager exists, yet. I'll share some more details soon for discussion.
Some remarks regarding the above:
- I would like to suggest including another identifier for the group/company as originator of the package, e.g. `groupId`. The combination of groupId and packageName (e.g. `acmecorp.fooService`) would be the unique id. This in my opinion would support a growing ecosystem nicely as reputation can be accumulated under a group name.
[...]

PR #5, opened 2016-10-04
add minifest-spec

pipermerriam:
Conversion of Tim's #2 into a pull request so we can discuss individual parts.


pipermerriam:
Don't merge this yet.  The idea is to move discussion into this PR so that we can discuss specific parts of the proposed spec.

pipermerriam:
This is potentially a [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) but I'm thinking that there should be a `manifest_version` so that future improvements to the package manifest specification can be accounted for.  Without this a client will need to accounting for all variations between versions of the spec in order to be compatible with all of them.

mhhf:
+1 for manifest_version
We should thing about extendability from the beginning, since the space and its requirements are changing rapidly.

manifest-spec.md
```diff
@@ -0,0 +1,37 @@
...
+1. Package manifests are primarily for non-contract actors and will not be consumed by contracts themselves.
+2. Following from that, we will likely store the manifest on some distributed protocol like IPFS or Swarm (although this is jumping ahead in the discussion).
+3. Since our off chain actors are primarily software, the message format needs to be easily readable by most programming languages.
+4. That said, it's likely human actors will write the manifest, meaning it needs to be in a format humans can easily understand and comprehend.
+5. Contrasting to the above, we need a format that takes up the least amount of space possible so that it costs the least to make available
+
+
+## Data to be stored within the manifest
+
+- package name
```
mhhf: https://gitter.im/ethpm/Lobby?at=57f56bf03c59573f6f0fd448
> Tim Coulter @tcoulter Okt. 05 23:09
> I’m contemplating we remove the name. That sounds like it might be registry level - lest the manifest name and the registered name not match.

I support this 👍

manifest-spec.md
```diff
@@ -0,0 +1,37 @@
...
+- author
+- version (preferably [semver](http://semver.org/))
+- description
```
mhhf: I think we should split this in a description (abstract), and a tags field.

manifest-spec.md
```diff
@@ -0,0 +1,37 @@
...
+- contract metadata
+  - abi
+  - unlinked binary
+  - deployed address if applicable
+  - name/address pairs for linked libraries
+  - compiler version for bytecode verification
+  - key/value pairs of the sha3 hash of events this contract can emit with the abi definition of that event (i.e., the contract metadata should contain information for all events that can be triggered by this contract, including those triggered by linked libraries)
```
mhhf: This should be omitted - as it can be reconstructed out of the other infos.

mhhf:
In dapple we found the need to store different environment information associated with a package/ deployed contract:
One simple use case could be a deployed Token contract stored together with its initial_balance which could look like this (admittedly very solidity centric):
```
...
"environment": {
  "token": {
    "type": "Token",
    "value": "0x67658a48be1b39db1c111dfca6f8fed09ac0da4e"
  },
  "initial_balance": {
    "type": "uint",
    "value": "0xffffff"
  }
}
```

should this be part of the manifest or solved on a different level?

pipermerriam:
I just pushed a pretty big update to this which attempts to define things a bit more precisely.

manifest-spec.md
pipermerriam: 👍 this is leftover from the previous version as I'm delaying on trying to define these parts.

manifest-spec.md
pipermerriam: 👍

manifest-spec.md
```diff
@@ -1,37 +1,140 @@
...
+
+### Contract Meta: `??-TODO-??`
+
+TODO: define this
+
+- contract metadata
+  - abi
+  - unlinked binary
+  - deployed addresses
+  - linked libraries
```
mhhf: >  - deployed addresses
>  - linked libraries

[I'm critical about this.](https://github.com/ethpm/epm/issues/8#issuecomment-253015071) I hope we can discuss it in detail.

manifest-spec.md
```diff
@@ -1,37 +1,140 @@
...
+The `version` field declares the version number of this release.  This value
+**should** be conform to the [semver](http://semver.org/) version numbering
+specification.  All manifests must **include** this field.
+
+* Key: `version`
+* Type: String
+* Package Manager Guide: Package Managers should validate this field conforms to the *semver* format and display a warning if it does not.
+
+
+### Description: `description`
```
mhhf: I'd suggest to add tags/keywords additionally to the description. Tags are the new links in the semantic web. Also every scientific paper provides keywords along the abstract.

manifest-spec.md
```diff
@@ -1,37 +1,140 @@
...
+
+The `package_name` field defines a human readable name for this package.  All
+manifests **should** include this field.
+
+* TODO: what should happen if this is missing (should it be required?)
+* Key: `package_name`
+* Type: String
+* Package Manager Guide: Package managers should use this field when registering new packages (TODO: define this better)
+
+### Author: `author`
```
mhhf: Author(s). In case there are several.

manifest-spec.md
```diff
@@ -1,37 +1,140 @@
...
+
+* `website`: Primary website for the package.
+* `documentation`: Package Documentation
+* `repository`: Location of the project source code.
+* TODO: what other common fields?
+
+* Key: `links`
+* Type: Hash(String: String)
+
+### Contract Meta: `??-TODO-??`
```
mhhf: Contract(s)
In case this is not a typo, why do you think they shouldn't be several contracts per package?

manifest-spec.md
```diff
@@ -1,37 +1,140 @@
...
   - unlinked binary
-  - deployed address if applicable
-  - name/address pairs for linked libraries
+  - deployed addresses
+  - linked libraries
   - compiler version for bytecode verification
-  - key/value pairs of the sha3 hash of events this contract can emit with the abi definition of that event (i.e., the contract metadata should contain information for all events that can be triggered by this contract, including those triggered by linked libraries)
 - external project URIs (homepage, repository uri, etc.)
 - dependencies (i.e., references to other packages and versions)
-- updated time (likely added by the packager)
```
mhhf: Where does the solidity source code live?

use-cases/README.md
```diff
@@ -0,0 +1,23 @@
+# Use Case Examples
+
+The following example packages can be used as test vectors for the supported
+use cases.
+
+
+## Base Contract
+
+These packages do provide any deployed instances of their contracts as they are
```
tcoulter: This should be "do not provide"

manifest-spec.md
tcoulter: My guess is external project URIs, which could be links to IPFS hashes, but we could include some other construct.

manifest-spec.md
tcoulter: :+1: `authors`

manifest-spec.md
tcoulter: I think for version 0.0.1 we shouldn't attempt to have packages that can reference contracts on other chains (which means addresses in the contract are fine, and reference the same environment that this package is published to). This is to say that a) we can't solve every problem on the first pass, and b) if you want packages to reference other chains you can extend the spec for your specific package manager, or simply create a new registry of deployed packages on the second chain.

manifest-spec.md
```diff
@@ -1,37 +1,140 @@
...
+### Contract Meta: `??-TODO-??`
+
+TODO: define this
+
+- contract metadata
+  - abi
+  - unlinked binary
+  - deployed addresses
+  - linked libraries
+  - compiler version for bytecode verification
```
tcoulter: Interesting suggestion. This spec is also assuming one specific contract language (solidity) and one specific compiler (solc). We might want to make `compiler` an object which contains information of the form:

```
{
  name: "solc",
  version: "0.4.2-342342342.a",
  flags: "--optimized" // or whatever it is
}
```

manifest-spec.md
pipermerriam: @mhhf I added a new stubbed out section for source code.

manifest-spec.md
pipermerriam: Changed to `authors`

manifest-spec.md
```diff
@@ -1,37 +1,140 @@
...
+### Description: `description`
+
+The `description` field *should* be used to provide additional detail that may be relevant for the package.
+
+* TODO: should this include a suggestion to use a specific markup format like markdown?
+* Key: `description`
+* Type: String
+
+
+### Links: `links`
```
pipermerriam: Probably but I'm not exactly sure how.  Adding a TODO note for this to be figured out.

manifest-spec.md
pipermerriam: Changed to be plural.

manifest-spec.md
```diff
@@ -1,37 +1,140 @@
...
+
+* Key: `links`
+* Type: Hash(String: String)
+
+### Contract Meta: `??-TODO-??`
+
+TODO: define this
+
+- contract metadata
+  - abi
```
pipermerriam: Added

manifest-spec.md
pipermerriam: Added

use-cases/README.md
pipermerriam: Fixed

tcoulter:
Awesome work Piper. Thanks for doing this.

manifest-spec.md
mhhf: I revoke my decision, a user given package name which doesn't depend on anything is a good thing. It can be seen as the most important made-up keyword.

release-manifest-spec.md
```diff
@@ -0,0 +1,147 @@
...
+### Chain: `chain`
+
+The `chain` field declares the blockchain that the contract addresses apply to.
+The chain is defined by a hash with block numbers as keys and block hashes as
+values.  Both block numbers and hashes *must* be hexidecimal encoded.
+Convention is to include the genesis block under the key `0x00` and the latest
+observable block at the time of release.
+
+* Key: `chain`
+* Type: Hash(BlockNumber: BlockHash)
```
pipermerriam: In our call yesterday we discussed avoiding including anything registry related in this specification.  We will still need to validate that is viable over the next few weeks but it's a nice simplification.

Another argument against including any registry information in these documents is it would make it more difficult to use a private mirror of packages which is a common use case for companies who don't want to have dependencies on 3rd parties for their build system.  Being able to have registry data declared at the package manager level should allow this.

In order to be *sure* that we can still support verifiable builds, our plan is to ensure that both the *source* and *dependency* data can be verified by whatever link format we choose (such as including a checksum).

Does this satisfy your requirements?

VoR0220: I definitely agree with @AFDudley that we should implement something along the lines of the ChainID/URI scheme here...it is a very flexible approach and effective approach to working with multiple chains. I think the terms are being overly loaded here. I think the main point of contention is the ability to get EXACTLY what we are looking for in any chains context...which is all very doable with this. I think doing something along this line will enable solidity contracts to be ported over to areas like Rootstock (if for whatever reason they want to do that) or to permissioned chains, or to EVM forked projects. 

Alternatively as was suggested in the meeting on Wednesday, we could utilize IPFS to make this happen as its hashing could point to a specific tx in a specific block in a specific chain.

pipermerriam:
+1

package-manifest-spec.md
```diff
@@ -0,0 +1,156 @@
...
+### Links: `links`
+
+TODO: This could potentially not be necessary, or should be removed.  Needs discussion.
+
+The `links` field *should* be used to provide URIs to relevant resources
+associated with this package.  When possible, authors *should* use the
+following keys for the following common resources.
+
+* `website`: Primary website for the package.
+* `documentation`: Package Documentation
```
VoR0220: This is not necessary don't think. This should be done by the documentation generated from solidity code.

pipermerriam: @VoR0220 this would be something like `https://web3py.readthedocs.io/en/latest/`.  A URI to a website or other resource that represents the formal documentation for the package.

release-lock-file-spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+
+## Filename
+
+Convention is to name this document `<version>.lock` where the `<version>`
+component is the full version string for the release.
+
+Package managers which are installing dependencies for development versions
+should keep their own version of this document under the lowercased name of the
+package manager such as `truffle.lock` or `dapple.lock`.  This file would
```
VoR0220: ++

release-lock-file-spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Manifest Version: `manifest_version`
+
+
+The `manifest_version` field defines the specification version that this
+document conforms to.  All release lock files **must** include this field.
+
+* Key: `manifest_version`
+* Type: Integer
+* Allowed Values: `1`
```
VoR0220: SemVer this.

VoR0220: Never mind you already said this.

release-lock-file-spec.md
```diff
@@ -0,0 +1,188 @@
...
+    * Format: Valid contract name matching regular expression `[_a-zA-Z][_a-zA-Z0-9]*]`
+* `address`:
+    * Required: No
+    * Type: String
+    * Format: Hex encoded ethereum address of the deployed contract.
+* `bytecode`:
+    * Required: No
+    * Type: String
+    * Format: Hex encoded unlinked bytecode for the compiled contract.
+* `runtime_bytecode`:
```
VoR0220: What is the difference between runtime_bytecode and normal bytecode? I also strongly believe that at the very least the ABI needs to be required. Quite possibly the bytecode too but not sure about that.

release-lock-file-spec.md
```diff
@@ -0,0 +1,188 @@
...
+    * Format: Combined [UserDoc](https://github.com/ethereum/wiki/wiki/Ethereum-Natural-Specification-Format#user-documentation) and [DevDoc](https://github.com/ethereum/wiki/wiki/Ethereum-Natural-Specification-Format#developer-documentation)
+* `compiler`:
+    * Required: Required if either `bytecode` or `runtime_bytecode` is present.
+    * Type: Hash
+    * Keys:
+        * `version`:
+            Type: String
+            Format: TODO
+        * `settings`: TODO
+* `link_dependencies`:
```
VoR0220: This may not need to exist as we might be able to find all of this from the metadata of a contract in future versions.

package-manifest-spec.md
VoR0220: mmmm....yea I'm not sure this be included in the package manager, but its in the area that is up to interpretation so I'm okay with leaving it here so long as that is explicitly stated.

release-lock-file-spec.md
pipermerriam: Can you expand on this?  I don't understand.

release-lock-file-spec.md
tcoulter: > runtime_bytecode is the unlinked version of the code that you should expect to find at the address of a deployed instance of this contract (ie, what is returned by eth_getCode.)

`eth_getCode` doesn't returned unlinked bytecode. Instead, it returns code that does not include a preamble; that is to say when you add a contract onto the network, the bytecode you're sending includes code that tells the vm to store your code somewhere (i.e., create a new contract).  `eth_getCode` only returns what was stored at an address by that preamble. (Aside: the preamble is standard, so you can add the binary for the preamble to anything returned by `eth_getCode` to get the code that was actually sent to the network.)

Regardless, I think it's important we include the unlinked bytecode which includes things like `__SomeContract___________________` inside the bytecode to ensure we can relink packages as we see fit. (Although I suppose we could always recompile...)

release-lock-file-spec.md
VoR0220: Basically the metadata output from a contracts bytecode will compile to a json output that contains all linked dependencies and remappings and such to them.

VoR0220: https://github.com/ethereum/solidity/pull/645

release-lock-file-spec.md
VoR0220: I think what you're trying to achieve with `runtime_bytecode` is be able to verify that the state is as it should be at a certain time, correct @pipermerriam ? If so I think that's incredibly valuable and the description should be updated to note this difference. It should be stressed that this is meant for safety and is not recommended for beginners use.

release-lock-file-spec.md
VoR0220: Ah wait...if this is for linking in terms of Libraries, then yes this may in fact need to happen.

pipermerriam: Yes, this is meant to be used for library linking.

package-manifest-spec.md
```diff
@@ -0,0 +1,156 @@
...
+## Filename
+
+Convention is to name this document `epm.json` which is short for *'ethereum
+package manifest'*.
+
+
+## Document Specification
+
+The following fields are defined for the package manifest.  Custom fields may
+be included.  Custom fields **should** be prefixed with `x-` to prevent name
```
axic: **must**?

It is probably worth explaining the process of adding fields to the specification.

package-manifest-spec.md
```diff
@@ -0,0 +1,156 @@
...
+
+
+* Key: `authors`
+* Type: List of Strings
+
+
+### Version: `version`
+
+The `version` field declares the current version number of this package.  This value
+**should** be conform to the [semver](http://semver.org/) version numbering
```
axic: I would argue it should must conform to semver.

axic:
Is it worth adding a license field (SPDX) to the manifest?

release-lock-file-spec.md
```diff
@@ -0,0 +1,188 @@
...
+document lifecycle.
+
+1. Release lock files are intended to be generated programatically by package management software as part of the release process.
+2. Release lock files will be consumed by package managers during tasks like installing package dependencies or building and deploying new releases.
+3. Release lock files will typically **not** be stored alongside the source, but rather by package registries *or* referenced by package registries and stored in something akin to IPFS.
+
+
+## Format
+
+The canoonical format for the package manifest is a JSON document containing a
```
axic: -> canonical

README.md
```diff
@@ -1,27 +1,49 @@
-# epm
-
-Ethereum Package Managment
+# EPM
 
 [![Join the chat at https://gitter.im/ethpm/Lobby](https://badges.gitter.im/ethpm/Lobby.svg)](https://gitter.im/ethpm/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
-# Goals
+Ethereum Packaging Specification
```
VoR0220: Recommend we change this to EVM Packaging Specification as it relates to more chains than just Ethereum.

README.md
```diff
@@ -1,27 +1,49 @@
...
-The goal of this repository is to create a specification that can be used for
-package management for the Ethereum Virtual Machine (EVM).
+A string matching the regular expression `[_a-zA-Z][_a-zA-Z0-9]*`
 
-This explicitely includes:
 
-* Specification for a *manifest* file.
+### Package Name
 
+A string matching the regular expression `[a-zA-Z][-_a-zA-Z0-9]*`
```
VoR0220: Still would like to recommend that we add a `Group ID` or `Repository` field into this.

package-manifest-spec.md
pipermerriam: If people want to use whole number versioning I don't want to stop them.  I agree that it's good for people to use this standard but there are lots of legitimate projects that don't use semver and I'm hesitant to put this as a requirement.

release-lock-file-spec.md
pipermerriam: * Def: **Canoonical:** *A humorous action performed in a canoe at or right around 12 Noon.*

README.md
pipermerriam: A bit more verbose but it could also be *"Ethereum Smart Contract Packaging Specification"* which may have more meaning to more people and still convey the same intent.  Or *"EVM Smart Contract Packaging Specification"*

package-manifest-spec.md
pipermerriam: This is another area where I don't want to be extra strict.  If people want to use other keys I don't see why the spec should stand in their way.

pipermerriam: > It is probably worth explaining the process of adding fields to the specification.

I don't think we actually have a process yet...

pipermerriam:
added `license` field.  Fixed *canoonical*.

README.md
pipermerriam: For metadata purposes with have `authors`.  For resolving packages I continue to push for keeping this out of the spec and leaving that up to package indexes.

Can you explain the use case for this again and why you think it belongs in these documents?

README.md
VoR0220: I want to stress that this is for more than Ethereum. It is for the EVM. We can go with either or after the change from Ethereum to EVM is made.

README.md
VoR0220: Interoperability...ah wait yes, we did have this argument about this and decided that this would be able to work under ENS, correct?

README.md
mhhf: > EVM Smart Contract Packaging Specification
👍 
Since indeed the use of packages could be used by different EVM chains like Rootstock

README.md
```diff
@@ -1,27 +1,38 @@
...
+        * Case for Package
+            * Defines the *current* version of the package.
+        * Case for Release
+            * Each release should have a clearly defined version.
+        * Case for Both
+            * Seems to work well for release workflow. Package manager would
+              populate the release manifest version field from the package
+              manifest version when creating a new release.
+* Validation of the concept of Package Dependencies vs Release Dependencies
+    * When a dependency is defined at the package level it *can* be either pinned to a specific release, or possibly a range of releases such as `^0.4.0` to allow any version in the `0.4.x` line.
```
AFDudley: >Package manager uses whatever registrar(s) it is configured to use to resolve all relevant version numbers that match the given semver. (if the value isn't a valid semver it probably just looks for an exact string match).

The problem with this is you're not conflating two different packages. The same package name and version from two different repos is not the same package.

>Package manager picks the greatest version that matches the specified semver number for the dependency.

Although semver helps with this, you're exploding the complexity of verifying the contract graph with this decision. It's better to keep the graph being verified small by defining its nodes as explicitly as possible. We are concerned with validating DAGs of contracts, right?

>I'm not sure I follow your logic at all here because my experience with package indexes is that they keep references to every release that's ever been made. I'm not sure why this would be any different.

For example, FreeBSD does a "ports freeze" before release to insure that the ports tree used for the production release is stable. (https://www.freebsd.org/doc/en_US.ISO8859-1/articles/committers-guide/ports.html#ports-qa-freeze-what)
Given that one of the major goals of this process is to verify that a package will perform as advertised in an adversarial environment, we need to support the ability to provide verified (and versioned) repos. I think we discussed this on gitter or something?

release-lock-file-spec.md
nmushegian: > Regardless, I think it's important we include the unlinked bytecode which includes things like __SomeContract___________________ inside the bytecode to ensure we can relink packages as we see fit. (Although I suppose we could always recompile...)

@tcoulter yes!! this is how dapple linking already works, originally we hoped to totally avoid addresses in contract constructors with this

README.md
pipermerriam: #### Semver vs Allowing * 

I understand that allowing non-semver version numbering adds complexity to package manager implementations but I'm still not comfortable *requiring* semver since I know of plenty of major software packages that use alternate versioning schemes.

#### Free BSD Thing

I understand what you are talking about with your FreeBSD example.  I just don't understand at all how it applies to this specification.

#### Pinning Dependencies.

I understand that two packages with the same version and name on two different package indexes *can* be different packages.  The current specification allows for un-ambiguously pointing at a specific release via an IPFS URI.  Would it appease your concerns if in the cases that a dependency was declared as a package_name/version pair that we supported some sort of content-hash as well so that once the dependency has beel resolved a package manager can then verify that the resolved package is the expected package?

AFDudley: >I understand that allowing non-semver version numbering adds complexity to package manager implementations but I'm still not comfortable requiring semver since I know of plenty of major software packages that use alternate versioning schemes.

I actually don't care if it's semver or not. The issue is your proposal allows for two different packages to resolve to the same identifier for no practical reason.

>I understand what you are talking about with your FreeBSD example. I just don't understand at all how it applies to this specification.

Again, just trying to get *global* package identifiers

>The current specification allows for un-ambiguously pointing at a specific release via an IPFS URI.

I think a blockchain URI makes more sense than an IPFS one. I think relying on IPFS in the spec is kinda lazy, We should make the content-addressing hash pluggable.

>Would it your stance on this if the spec was modified such that in the cases that a dependency was declared as a package_name/version pair that we supported some sort of content-hash as well so that once the dependency has beel resolved a package manager can then verify that the resolved package is the expected package?

I think a BIP122 URI is better than a protocol-dependent content-hash but a protocol independent content-hash is at least as good as a BIP122 URI if not significantly better. I understand Swarm has adopted IPFS' content-hashing so if the spec is:

`<package_name>==<semver>#<pluggable_content_hash>`

this makes sense to me. But mixing hashes and URIs seems sort of silly. either map the global URIs to the content-hashes or just use content-hashes. I think we should just refer to all external objects via URIs. or things that are trivial to convert to URI, content-hashes have this property, but they aren't human readable, I think putting the mapping from human readable to content-hash in this file is kinda weird... but I'd need to think about it more to be sure.

VoR0220: The thing is, apparently, we can create a blockchain URI with IPFS. Now granted, perhaps this is relying too much on IPFS, but it's a fairly solid protocol. There may also be room for a combination approach...But I'm not sure how we want to go about this...As for Swarm + IPFS, there is a difference in the headers, but I believe that could be added accordingly if they are there, we just need a way to get that one hash and then have the headers added subsequently.

AFDudley: Sorry, the above is a bit confused. My over-arching goal is to make it easy to say "These packages have been verified and tested together". Part of achieving that goal is mapping the name of a package to the source and byte code of a package, such that there is no ambiguity to the people reading it. Content-addressing half helps, because I can't assume that IPFS will be available just because a customer is using the EVM. Also, traditional URIs and content-addressed URIs are both URIs so any place where we are specifying a content-address we could more generally specify a URI.

VoR0220: Either way, we should be able to map to both traditional centralized servers as well as DHTs.

package-manifest-spec.md
```diff
@@ -0,0 +1,167 @@
...
+
+
+### Dependencies: `dependencies`
+
+the `dependencies` field defines a key/value mapping of ethereum packages that
+this project depends on.
+
+* All keys **must** be valid package names matching the regular expression `[a-zA-Z][-a-zA-Z0-9_]*`
+* All values **must** conform to *one of* the following formats:
+    * IPFS URI:
```
AFDudley: Seems like this could reference a file that maps to a different chain? This should only specify names.

tcoulter:
Any chance we can merge this and talk about minor changes as needed? A lot of what's in discussion right now can be easily changed.

tcoulter:
That is to say I'd prefer each one of these comments to be its own discussion.

tcoulter:
Merged. Hopefully I didn't derail discussions currently in progress. Please bring up new discussions as individual issues. Thanks guys!

PR #10, opened 2016-11-16
Use relative paths on source listings

tcoulter:
Couple things:

1. I don't think it makes sense to allow absolute paths here, as that makes package manifests environment specific (something we don't want to encourage);
1. If you assume the above to be true, this means we want relative paths -- I'd argue these paths should be relative to the location of the package manifest (`epm.json` file).

Technically someone could still use `../../../` etc. to get out, but at least we know how to resolve those if it's relative to the `epm.json` file.

### DIFF
--- a/package-manifest-spec.md
+++ b/package-manifest-spec.md
@@ -127,13 +127,13 @@ following keys for the following common resources.
 ### Source Files: `sources`
 
 The `sources` field defines a set of source files that comprise the source code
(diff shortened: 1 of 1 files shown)


tcoulter:
@pipermerriam your changes look good to me!

pipermerriam:
Would like to merge this if there are no objections.

tcoulter:
:shipit:

PR #11, opened 2016-11-16
Propose change of address to Contract Instance Object.

VoR0220:
Addresses should be an array and be updated for every time there is a deployment I think. Or at the very least leave the option open for such a package if that is desired.

### DIFF
--- a/release-lock-file-spec.md
+++ b/release-lock-file-spec.md
@@ -145,8 +145,8 @@ A *Contract Instance Object* is an object with the following key/values.
     * Format: Valid contract name matching regular expression `[_a-zA-Z][_a-zA-Z0-9]*]`
 * `address`:
     * Required: No
-    * Type: String
-    * Format: Hex encoded ethereum address of the deployed contract.
+    * Type: []String (Array of Strings)
+    * Format: Hex encoded ethereum address(es) of the deployed contract. If null, assume not deployed. 
 * `deploy_transaction`:
     * Required: No
     * Type: String


tcoulter:
Aren't past addresses stored forever in old lockfiles? Each time you publish a new package, if you've redployed, the new address will exist in the lockfile.

VoR0220:
I suppose that could be the case, but even so, say I have a factory and deploy numerous instances of said contract from a certain contract with only an enum changed that effects the functionality of the entire contract. In this instance we would want those addresses recorded, would we not?

tcoulter:
Good point. Perhaps `address` should be `addresses`?

VoR0220:
or `address(es)`

tcoulter:
I don't like ambiguity -- this is mostly a machine readable doc, so an `addresses` array with a single entry is perfectly fine. It's extra work if it can be a string or an array. If this were the manifest I might argue differently; but this is for the lockfile so we might as well constrain the possibilities.

tcoulter:
Sorry, I think my response sounds strong via text. Not trying to just shut you down.

pipermerriam:
This case is covered.  The `contracts` key would look something like this.

```javascript
{
  ...
  'contracts': {
    'MKRToken': {
      'contract_name': 'StandardToken',
      ...,
    },
        'DAIToken': {
      'contract_name': 'StandardToken',
      ...,
    },
  },
}
```

VoR0220:
@pipermerriam that sounds like more to parse and that's the problem. That'll bloat the file.

pipermerriam:
Was there resolution on this?  I think we decided to stick with the current spec.  If that is the case can we close this?

VoR0220:
close for now yes. I think we definitely should revisit this at a later time.

PR #13, opened 2016-11-17
Rename manifest_version to lock_file_version

pipermerriam:
This renames the field in the Release Lock File from `manifest_version` to `lock_file_version`.  Open to a better name if anyone can think of one.


tcoulter:
Good change.

PR #14, opened 2016-11-18
Specify compiler format and use Object instead of Hash

pipermerriam:
* Changes from using the term `'Hash'` to using `'Object'` everywhere.
* Defines the `compiler` key in the Contract Instance Object definition.


tcoulter:
:+1: from me.

VoR0220:
Yup 👍

PR #15, opened 2016-11-18
remove contracts key from package manifest

pipermerriam:
Removes the `contracts` key from the package manifest.

Reasoning is that I am failing to see how this key is actually useful.  I originally thought it would be used by package managers when building a release but now that I'm actually building a release I am finding that these values aren't helpful.


tcoulter:
:+1:

VoR0220:
This is useful whenever deciding to generate instances for a certain source...But I don't think this is probably the way to do it.

pipermerriam:
Closing.  I did this in #23

PR #16, opened 2016-11-18
change suggested name document name for release lock file

pipermerriam:
Changes the suggested name of the release lock file to `<version>.json` instead of `<version>.lock` since it is a json document and having the `json` extension seems more appropriate in this case.

### DIFF
--- a/release-lock-file-spec.md
+++ b/release-lock-file-spec.md
@@ -24,13 +24,13 @@ single JSON object.
 
 ## Filename
 
-Convention is to name this document `<version>.lock` where the `<version>`
+Convention is to name this document `<version>.json` where the `<version>`
 component is the full version string for the release.
 
 Package managers which are installing dependencies for development versions
-should keep their own version of this document under the lowercased name of the
-package manager such as `truffle.lock` or `dapple.lock`.  This file would
-typically be excluded from version control.
+should keep their own version of this document. Convention is to use the
+lowercased name of the package manager such as `truffle.lock` or `dapple.lock`.
+This file would typically be excluded from version control.
 
 
 ## Document Specification


tcoulter:
Generally into this. I haven't yet saved the lockfile to disk yet -- I've only ever downloaded it and kept it in memory. But I'm fine with the naming scheme.

VoR0220:
How will you differentiate the name between an epm.json and an epm.json? It would seem to me that the .lock file is only a name. I vote no.

pipermerriam:
@VoR0220 I don't think this is an issue.

* Package Manifest: `epm.json`
* Lockfile: `<version>.json`

No conflict unless someone uses the string `'epm'` as their version *and* they actually keep that file locally in the source repo.

VoR0220:
<Version>.json? I'm going to come out strongly against this because it isnt consistent in what to look for.

pipermerriam:
> because it isnt consistent in what to look for. 

@VoR0220 I don't understand what you mean by this.

VoR0220:
There is no consistent way to define what is and what isnt a lock file.

tcoulter:
Thought more about the naming scheme. I do like changing the extension to `.json`. I don't like prefixing that with the version, however. Instead, I'll be saving lockfiles as `lock.json`. Reasoning is:

* There is only ever one lockfile in any installed package directory, so uniqueness isn't required.
* Naming it x.x.x.json is inconsistent, making it harder to read from as you can't guarantee its name.
* Removing x.x.x doesn’t remove available data as the version exists within the lockfile.

pipermerriam:
@tcoulter I think you are conflating the two lock files which is understandable because our names are confusing.

1. A file that is generated on each release which we've been referring to as the **Release Lock File**.  This is the file that this PR is targeted at renaming.
2. A file that each package manager uses to track the current state of installed packages.  This file is not *currently* specified.

I think you are talking about the latter of these, in which case this PR doesn't apply to the naming scheme.

pipermerriam:
Also, see https://github.com/ethpm/epm-spec/issues/22

tcoulter:
On my phone, so will be brief. I'm talking about the first one.

A requirement that we save this file on disk the same way is unnecessary.
It's an implementation detail that doesn't matter for the package
specification.

I'll be saving my packages like `installed_contracts/<package name>/<lock
file name>`, for now, until dependencies get more complex (*subject to
change). For me, it's easier to call the lock file `lock.json` so I can
easily find it later when searching what's installed.


On Dec 4, 2016 7:52 AM, "Piper Merriam" <notifications@github.com> wrote:

Also, see #22 <https://github.com/ethpm/epm-spec/issues/22>

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/ethpm/epm-spec/pull/16#issuecomment-264711605>, or mute
the thread

VoR0220:
If we're going to name it like that, perhaps an epmlock.json works better? I think making a lock version for each release is redundant and the version can be and is specified in the lock file itself.

tcoulter:
I don't think you need to follow my naming scheme at all. There's nowhere
in the package spec that governs this (or needs to). You should save this
file on disk however you like.

PR #18, opened 2016-12-01
adding json-schema and document generation

mhhf:
This branch includes a first json-schema draft of `package-manifest-spec` and `release-lock-file-spec`. It also includes an automatic json-schema to markdown documentation generator which unfortunately don't captures all available information, therefore the original documentation are kept in place, although I would argue to deprecate them as soon as a sufficient documentation generator is found to reduce redundancy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "epm-spec",
+  "name": "ethpm-spec",
   "version": "1.0.0",
   "description": "Ethereum Package Manager Specifications",
   "main": "index.js",
--- a/package.json
+++ b/package.json
@@ -1,0 +1,29 @@
+{
+  "name": "epm-spec",
+  "version": "1.0.0",
+  "description": "Ethereum Package Manager Specifications",
+  "main": "index.js",
+  "scripts": {
+    "build": ""
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/ethpm/epm-spec.git"
+  },
+  "keywords": [
+    "ethereum",
+    "package",
+    "manager",
+    "epm",
+    "ethpm"
+  ],
+  "author": "",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/ethpm/epm-spec/issues"
+  },
+  "homepage": "https://github.com/ethpm/epm-spec#readme",
+  "dependencies": {
+    "json-schema-to-markdown": "^1.0.3"
+  }
+}
--- a/spec/release-lock-file.schema.json
+++ b/spec/release-lock-file.schema.json
@@ -49,7 +49,8 @@
       "items": {
         "title": "URI",
         "description": "must be valid BIP122 URIS",
-        "type": "string"
+        "type": "string",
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "epm-spec",
```
tcoulter: I'm trying to get everything renamed to `ethpm` to reduce any name collisions. Can we rename to `ethpm-spec`? This should likely coincide with a renaming of the repo as well.

spec/release-lock-file.schema.json
```diff
@@ -0,0 +1,134 @@
...
+    },
+    "chain": {
+      "title": "Chain",
+      "type": "array",
+      "description": "The `chain` field defines the blockchain that should be used for any addresses provided with this package.  A blockchain is defined using [BIP-122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki).  A matching blockchain is one on which all resources defined by the list of BIP122 URIs can be found. *If* this release lock file includes any addressed contracts this field **must** be present.  Convention is to define a chain using a single URI which points to the latest observable block hash and includes the genesis hash as the `chain_id`.",
+      "default": [],
+      "items": {
+        "title": "URI",
+        "description": "must be valid BIP122 URIS",
+        "type": "string"
```
pipermerriam: can we add `"format": "uri"` to this and any other URI formatted properties?

See http://json-schema.org/latest/json-schema-validation.html#rfc.section.7

spec/release-lock-file.schema.json
```diff
@@ -0,0 +1,134 @@
...
+      "type": "object",
+      "description": "The `contracts` field declares information about the deployed contracts included within this release.",
+      "patternProperties": {
+        "^[a-zA-Z][-a-zA-Z0-9_]*$": {
+          "type": "object",
+          "description": "A *Contract Instance Object* is an object with the following key/values.",
+          "properties": {
+            "contract_name": {
+              "type": "string",
+              "description": "Valid contract name matching regular expression `[_a-zA-Z][_a-zA-Z0-9]*]`"
```
pipermerriam: Can we add `"pattern": "^[a-zA-Z][-a-zA-Z0-9_]*$"` to this to enforce the contract naming scheme.

http://json-schema.org/latest/json-schema-validation.html#rfc.section.5.8

package.json
mhhf: :+1:

spec/release-lock-file.schema.json
mhhf: :+1:

spec/release-lock-file.schema.json
mhhf: :+1:

pipermerriam:
I'm comfy merging this as is.  I don't think we're shooting for perfection right now and I'd like to go ahead and start using these to validate the manifests/lock-files that I'm generating.

mhhf:
👍  go for it!
apart from writing the spec, I didn't tested it against example files yet, so expect the validation to break.

PR #19, opened 2016-12-02
add schema version and change sources from array to object

mhhf:


spec/release-lock-file.schema.json
```diff
@@ -32,12 +33,14 @@
...
-      "items": {
-        "title": "Source Path",
-        "type": "string",
-        "description": ": Strings must be formatted as valid filesystem paths. All paths must be relative paths beginning with ./. All paths must resolve to to a path located within the base directory of the package (i.e., the location of epm.json"
+      "patternProperties": {
+        ".*": {
```
VoR0220: I'm beginning to think that this source path key may be unnecessary. Especially with the json output from solidity coming out and the fact that we're pointing to where to get the source code which can tell us the source paths so we can figure it out natively.

mhhf: @VoR0220 Lets stick to the `./` notation atm and discuss a protocol update once the json output from solidity is ready. Also mind, that we are trying to do a EVM spec,  not a solidity spec, so we should think about the general case here.

spec/release-lock-file.schema.json
```diff
@@ -2,6 +2,7 @@
...
   "type": "object",
+  "version": 1,
```
VoR0220: Because semver

pipermerriam:
I'm 👍 to merge this.

PR #23, opened 2016-12-05
Piper/update specs and schemas

pipermerriam:
* Removes the auto-generated specs
* Sets all `descriptions` in the JSON-schema documents to `'TODO'`
* Denote whether fields are required in the package manifest.
* Remove the `contracts` key from the package manifest.
* Change wording of filename suggestion for Release Lock File
* Add `pattern` to JSON-schema for package name
* Add `build_dependencies` to Release Manifest JSON-schema
* Add `enum` to spec versions
* Add `enum` to compiler types in JSON-schema
* Set `contract_name` in Contract Instance Object to be required.
*


mhhf:
:+1:

PR #24, opened 2016-12-05
Define Entry Point for Package Manifest

VoR0220:


pipermerriam:
I still don't understand how this would be used.  Can you provide some use cases?

mhhf:
Can this be solved at a package manager level, like multi chain is done?
This PR would make EthPM solidity based, which we wanted to omit.

VoR0220:
good point @mhhf I keep forgetting that. Closing.

PR #26, opened 2016-12-05
update version of package-manifest spec

mhhf:
I remember the request that this should be a string rather a integer.


pipermerriam:
Why version 2?

mhhf:
i changed it, so i had to update the version.

tcoulter:
Are we at a decided upon version 1 yet?

pipermerriam:
I don't think we're at that point yet.  Can we agree that we'll all continue to update our tools until we have a releasable product and that will be called version `"1"`

pipermerriam:
Releasable product is probably defined as at least two of our frameworks being able to run

`thing install ipfs://Qm....` 

where the IPFS link there points at a Release Manifest and all dependencies are also direct links to Release Manifest files (i.e. we are not requiring any registrar related features).

PR #29, opened 2016-12-06
Disallow uppercase characters in package names

pipermerriam:
Don't allow uppercase characters in package names.

### DIFF
--- a/package-manifest-spec.md
+++ b/package-manifest-spec.md
@@ -138,7 +138,7 @@ the *Release Lock File*.
 the `dependencies` field defines a key/value mapping of ethereum packages that
 this project depends on.
 
-* All keys **must** be valid package names matching the regular expression `[a-zA-Z][-a-zA-Z0-9_]*`
+* All keys **must** be valid package names matching the regular expression `[a-z][-a-z0-9_]{0,214}`
 * All values **must** conform to *one of* the following formats:
     * IPFS URI:
         * The resolved document **must** be a valid *release lock file*.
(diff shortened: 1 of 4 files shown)


tcoulter:
👍

VoR0220:
No

pipermerriam:
@VoR0220 I merged this due to the 3:1 vote.  I am still curious to hear your *why* but without knowing the *why* I chose to go ahead and move forward with this.  Still of course open for discussion.

VoR0220:
because if we want a mixed case, that could be useful. Also enables more differentiation.

tcoulter:
I don't think we should allow people to make a distinction between `monax-stdlib`, `Monax-stdlib`, and `monax-STDlib`. Actually, thinking about it we probably don't want underscores unless we want to allow people to create `monax-stdlib` and `monax_stdlib`. We have enough naming issues already.

I vote no uppercase, no underscore.

VoR0220:
so I agree with not wanting to make the distinction, but I think mixedCase is useful.

tcoulter:
Right, but how is it useful given my examples above of how it's not useful?

tcoulter:
Reducing the namespace (character space) is probably better for everyone. e.g., http://bgr.com/2016/11/21/fake-google-vitaly-popov-scam-spam/

tcoulter:
As an aside, is this something we can actually enforce?

VoR0220:
I suppose from the package manager level. But its kind of like an agreement we all make to do this. Not really enforceable for all newcomers who think our package managers suck :p

PR #30, opened 2016-12-06
added spec for chains file and modified lock file

VoR0220:
Problem: Currently we have chain related information inside the lock file. This is problematic because the lock file is meant to be a release file, and it shouldn't be updated once it's released. But say that I deploy my contract on a new chain after the release and want to keep that information somewhere that it was in fact deployed. This requires us to update the lock file...which is bad practice for a release.

Solution: Separate dependencies from linking. Linking can be done either by deploying a copy of contract libraries at new addresses or by simply referencing this chains file. The only problem I'm forseeing is how do we enable it to be dynamically updated and still easily trackable while also keeping it decentralized. Indeed a very difficult problem methinks. Anyways. Let me know what you think. I feel pushback a coming. 

Signed-off-by: VoR0220 <rj@erisindustries.com>

### DIFF
--- a/chains-file-spec.md
+++ b/chains-file-spec.md
@@ -1,0 +1,97 @@
+# Chain File Specification
+
+This document defines the specification for the **Chain File**.  The
+chain file is a dynamically updated file holding information about where contracts are
+deployed at on different chains. This is an optional file to download, it is meant purely 
+to save the cost of redeployment. In the case that it is present, it should be a hidden file. 
+
+** Problem: How do we dynamically update this file from multiple parties? How do we ensure its integrity? How do we prevent spam? **
+
+## Guiding Principles
+
+The chain file specification makes the following assumptions about the
+document lifecycle.
+
+1. Chain files are intended to be generated programatically by package management software as part of the deployment process.
+2. Chain files will be consumed by package managers during tasks like linking and deploying packages containing libraries or contracts of particular interest.
+3. Chain files are dynamic whereas the manifest and lock files are static. This is to be updated every time there is a new deploy of the package onto a release chain to be actually utilized by multiple parties.
+
+
+## Filename
+
+When deploying, if it exists, the referenced file should be a hidden `.packageName-chains.json` file where `packageName` denotes the name of the package. Hidden because we don't want to spam this thing to hell.
+
+## Format
+
+The canonical format for the chain file is a JSON document containing a
+single JSON object.
+
+### Chain File Version: `chain_file_version`
+
+The `chain_file_version` field defines the specification version that this
+document conforms to.  All chain files **must** include this field.
+
+* Key: `chain_file_version`
+* Type: Integer
+* Allowed Values: `1`
+
+### Release File Hash: `release_hash`
+
+The `release_hash` field defines a keccak256 hash of the release lock file that this document conforms to.
+All chain files **must** include this field.
+
+* Key: `release_hash`
+* Type: String
+
+
+The following fields reference a blockchain as defined using
+a subset of the [BIP-122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki) spec, specified below:
+
+```
+<chainID>://<genesis_hash>/block/<latest block hash>
+```
+
+We will break this into portions
+### ChainID: `chainID`
+
+The `chainID` field defines the blockchain that should be used for any addresses
+provided with this package. This field **must** be present.
+* Key: `chainID`
+* Type: String mapping to list of genesis hashes
+
+### Genesis Hash: `genesis_hash`
+
+The `genesis_hash` represents the blockhash of the first block on the chain. This field **must** be present.
+* Key: `genesis_hash`
+* Type: string mapping to list of `contract` strings
+
+### Contract Object: `contract`
+
+A contract object containing information on addresses and deploy times. Similar to contract instance object in lock file.
+* Key: `contract`
+* Type: String mapping to list of deploy objects
+
+### Deploy Objects: `deployed`
+An object containing an `address`, and either a `deploy_block` or `deploy_transaction`.
+* Key: `deployed`
+* Type: Object
+
+* `address`:
+    * Required
+    * Type: String
+    * Format: Hex encoded ethereum address of the deployed contract.
+* `deploy_transaction`:
+    * Required if `deploy_block` not specified
+    * Type: String
+    * Format: [BIP122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki) URI which defines the transaction in which this contract was created.
+* `deploy_block`:
+    * Required if `deploy_transaction` not specified
+    * Type: String
+    * Format: [BIP122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki) URI which defines the block in which this contract was created.
+
+### Latest Block Hash: `latest_block_hash`
+
+`<latest block hash>` represents the hash of the latest block that's been reliably confirmed. This is not required but is strongly recommended.
+
+*If* this release lock file includes any addressed contracts
+this field **must** be present. 
--- a/release-lock-file-spec.md
+++ b/release-lock-file-spec.md
@@ -11,9 +11,9 @@ The release lock file specification makes the following assumptions about the
 document lifecycle.
 
 1. Release lock files are intended to be generated programatically by package management software as part of the release process.
-2. Release lock files will be consumed by package managers during tasks like installing package dependencies or building and deploying new releases.
+2. Release lock files will be consumed by package managers during tasks like installing package dependencies.
 3. Release lock files will typically **not** be stored alongside the source, but rather by package registries *or* referenced by package registries and stored in something akin to IPFS.
-
+4. Release lock files are static. Once the release is in, they **should not** be updated further.
 
 ## Format
 
@@ -90,26 +90,6 @@ Sources are declared in a key/value mapping.
 * Type: Object (String: String)
 
 
-### Chain: `chain`
-
-The `chain` field defines the blockchain that should be used for any addresses
-provided with this package.  A blockchain is defined using
-a subset of the [BIP-122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki) spec, specified below:
-
-```
-blockchain://<genesis_hash>/block/<latest block hash>
-```
-
-The `<genesis hash>` represents the blockhash of the first block on the chain, and `<latest block hash>` represents the hash of the latest block that's been reliably confirmed (package managers should be free to choose their desired level of confirmations).
-
-*If* this release lock file includes any addressed contracts
-this field **must** be present. 
-
-* Key: `chain`
-* Type: String
-* Format: `blockchain://[0-9a-fA-F]{64}/block/[0-9a-fA-F]{64}`
-
-
 ### Contracts: `contracts`
 
 The `contracts` field declares information about the deployed contracts
@@ -147,18 +127,6 @@ A *Contract Instance Object* is an object with the following key/values.
     * Required: Yes
     * Type: String
     * Format: Valid contract name matching regular expression `[_a-zA-Z][_a-zA-Z0-9]*]`
-* `address`:
-    * Required: No
-    * Type: String
-    * Format: Hex encoded ethereum address of the deployed contract.
-* `deploy_transaction`:
-    * Required: No
-    * Type: String
-    * Format: [BIP122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki) URI which defines the transaction in which this contract was created.
-* `deploy_block`:
-    * Required: No
-    * Type: String
-    * Format: [BIP122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki) URI which defines the block in which this contract was created.
 * `bytecode`:
     * Required: No
     * Type: String


pipermerriam:
> But say that I deploy my contract on a new chain after the release and want to keep that information somewhere that it was in fact deployed. 

Why wouldn't you just make a new release?

VoR0220:
We could do it that way, but that sounds like it's going to get messy really quickly.

mhhf:
how do you understand messiness here?

VoR0220:
Lets address the above. How is having 5k different `version`.jsons floating around good for a coherent package management system? I might be missing something but that doesn't sound right...if someone specifies that they want a package of x version, they should get exactly that.

tcoulter:
> How is having 5k different version.jsons floating around good for a coherent package management system?

For some perspective, `npm` has a separate lockfile for every release: https://registry.npmjs.org/truffle/

VoR0220:
@tcoulter that's not what I'm saying. There should be a separate lockfile for every release. What I'm saying is that once you do release, you should not be modifying said release lockfile, but due to the conditions of what we're working with in terms of chain deployment, that's what we're inviting and it's bad practice. Hence why we need to separate these parts.

tcoulter:
> What I'm saying is that once you do release, you should not be modifying said release lockfile, 

I don't see how this is even possible. If you change the lockfile, it's a new file and will have a different IPFS/Swarm hash. If changed, it *should* be associated with a new release, in which case your statement of "There should be a separate lockfile for every release." applies.

tcoulter:
I don't think we've gotten to the core of your issue and why you feel you need to change the lockfile.

pipermerriam:
ok, looking at your original post you say this.

> But say that I deploy my contract on a new chain after the release...

In that case, under the current specification you would be required to make a new release because  each lockfile contains a single chain definition via the chain URI.

tcoulter:
What if we were to move the chain definition to each contract object that has an address rather than for the whole package?

VoR0220:
@tcoulter What im suggesting is that we take out chain based information and separate that from the lock file. The lock file purely tracks file dependencies in this situation. Chains now get a new file where this can be updated freely next to the release. 

@pipermerriam suppose that i release a package with information onto chain a. Two weeks later somebody downloads my package and tries to deploy onto chain b. How does that developer export information to people interested in their project that the contract is deployed on chain b? We shouldnt modify the release. So whats the way to resolve this?

pipermerriam:
> suppose that i release a package with information onto chain a. Two weeks later somebody downloads my package and tries to deploy onto chain b. How does that developer export information to people interested in their project that the contract is deployed on chain b?

Under the current spec I see two paths for this.  I'm sure there are more.

1. They could reach out to the package author and ask them to publish a release on chain B
2. They could fork the package and deploy it themselves on chain B

Thoughts?

tcoulter:
My hunch was that what @VoR0220 was describing was a fork. Once a package is deployed to a different chain or registry, to me it's a different package.

VoR0220:
@pipermerriam what I'm saying is that the current spec is incorrect and needs to be changed. We need to think of how we blend the old style of package management under this new paradigm of coding. We need to take the unique values of this paradigm into account and adjust our spec for them.

tcoulter:
> Again I have to ask, what's wrong with the solution I am proposing in this PR itself? Is there something wrong with segregating chain information into its own file and having that file be updatable frequently?

In general all parts of a released package should be immutable. I think that's the main differentiator between our two different ways of thinking.

Also, a lockfile only references IPFS information via URIs, and if the code didn't change, then the code won't contribute to added bloat as IPFS will see it as the same file and give it the same hash regardless of the chain it has been deployed on. The extra lockfile might be extra bloat, but miniscule. 

That said, I don't like separating deployed address away from chain identification as an address is inextricably linked to its deployed chain. If I were you, I'd not add any deployed addresses at all to your lockfiles and instead manage chain data as a separate feature of monax. Note that according to the spec, you don't need chain data if you haven't included any deployed addresses in your lockfile.

> If this release lock file includes any addressed contracts this field must be present.

tcoulter:
Additionally, there's nothing stopping you from adding `x-monax-chain-info` and referencing an IPFS document or somewhere else that supports your use case.

VoR0220:
So I agree that all parts of the release should be immutable. This is simply saying that the parts that are unlikely to remain immutable be segregated from the lock file. I suppose I could do it this way for our purposes. But I maintain here that this is the incorrect and an improper way of doing package management involving multiple chains.

nmushegian:
@tcoulter What about multiple instances of "the same" package deployed multiple times on the same chain? Do we have to create a thin wrapper package to give each one a distinct name?

tcoulter:
In my opinion a package is a mapping of (chain, registry, name, version) => lockfile. If anything on the left changes I'd consider it a new release/package.

mhhf:
@nmushegian this is covered by the spec as every deployed instance has a distinct "name" therefore all instances can fit into one lockfile

tcoulter:
As an edit to the above: If you don't include chain information (deployed addresses) in your package then the mapping becomes (registry, name, version) => lockfile. As @mhhf suggested you can always point items in different registries but on the same change to the same lockfile.

AFDudley:
@tcoulter
>In my opinion a package is a mapping of (chain, registry, name, version) => lockfile. If anything on the left changes I'd consider it a new release/package.

This is the deployed thing, not the thing written by the developer. all the silly terms this team has come up with I've forgotten, but the reason you have two files in the first place is so you can target different chains. The idea that the first file, the human readable file, also *somehow* dynamically pulls in the chain/registry information is perfectly sane; those are deployment and validation concerns not pure guy-writing-code concerns. (I'm super busy doing other stuff, hopefully I didn't hurt anyone feelings.)

VoR0220:
@tcoulter how is the same code a different package though? That befuddles me. If we want to have it in the manifest that you have to name the chain you're using too, I think this would be a workable compromise because then we can atleast index to the correct lock file if we go with your worldview of different chain to same code being a different package.

VoR0220:
and if this is the case the version naming needs changing for locks to distinguish for locks.

tcoulter:
No feelings hurt. I don't think the spec in its current form disallows your use case. 

An address isn't useful without chain information. You want chain information in a different file. I think you guys should put your chain information and addresses in a different file and be done with it. Otherwise we're mostly discussing the philosophy of a package -- something I could better describe with pictures than words.

Lets assume all code we're talking about lives on the same IPFS network. At that point all we're discussing is how many lockfiles to create and how many items to put in the registry, which is all a function of your individual package manager.

To put this conversation into perspective, if I deploy a single version of a JS library to both NPM and Bower, is it one package or two?

AFDudley:
>Lets assume all code we're talking about lives on the same IPFS network.

everything I do is based on their being multiple chains, some public, some private, I can't really make this assumption since sometimes the exact same contract (for a two-way peg for example) will be published in public and in private.

>To put this conversation into perspective, if I deploy a single version of a JS library to both NPM and Bower, is it one package or two?

it's the same JS code that runs and thus it should have the same version. Another example is code that is published as debian deb and an ubuntu deb has the same version associated with it, but the deb files *may not* have the same sha1

nmushegian:
> An address isn't useful without chain information.

@tcoulter Not sure I agree with this... you can determine address->chain (except for forks in chains without replay protection)

VoR0220:
@nmushegian you can? Do tell. This might actually be very useful information.

tcoulter:
@VoR0220 I assume he means that if an address exists on a chain, you can assume it to be the correct chain because a lot of things had to exist cryptographically to create that address (each address has a 1 in a bajillion chance of existence).

nmushegian:
@tcoulter that's right, I'm not 100% sure if the eth address definition has the properties we need, but it would be a real shame if it did not

pipermerriam:
After reading through all of this I haven't seen a use case that can't be covered by the current specification.

* In the case where there is need to deploy the same package on multiple chains a release lock file can be made for each chain.
* In the case where there are multiple deployed versions of the same package on the same chain, the current spec for the release lock file allows for this.

I'm going to vote 👎 on this addition on the following grounds:

* The use cases are doable under the current spec (though I understand that they may be cleaner with this addition)
* It adds complicates the specification for a use case that we previous have said is not part of our core goals, that is multi-chain deployments.  Note this doesn't mean the spec doesn't allow deployment on multiple chains but it intentionally doesn't have a concept of a release that spans multiple chains.
* Multi-chain deployments are also doable under the current spec but I'm not doing to go into how that work work in detail.  Basically, there is nothing that says a dependency can't point at a release on another chain.

VoR0220:
Not only that, but you could check the chain at that address and a hash of the opcodes themselves and see whether or not the assembly is the same as the assembly in the current repo.

j-h-scheufen:
Hi all, my apologies that I haven't been able to follow the spec development as close as I would've liked, but I want to add my 2 cents here as I think there are some misunderstandings and we're conflating two different types of "locks" needed.

1. The (traditional) "lock" is generated when the developer of a package certifies and publishes a release and pins it to the dependency versions that were used for compilation, therefore certifying compatibility. This does not include deployment or installation (on a chain), yet.
2. "Locking" (in the sense that it's used in this group) pins a package version to a particular chain and links to resources on that chain.

Lock 1) is what typically happens in any programming language where versioned packages are released. (It's mostly not called a "lock" as far as I can say, but is the release manifest).
Lock 2) in traditional software usually happens automatically and it's not persisted, e.g. when installing Linux packages or httpd mods or whatever, the dependencies to modules on the user's OS are resolved and linked adhering to the constraints set in Lock 1).
To me, the difference here is that our OS is a blockchain and the main reason why I'd want to persist the linking to a file is that it is useful to others and is worth publishing for people that want to use the same chain and link to my package on it.

I hope I'm making sense here (it's late and thoughts are not so fluid). I apologize in advance in case I'm not following things that were already specified, but here's how I see it:

- The reason why I call Lock 1) a "lock" instead of simply a manifest is that it does not suffice to just declare a dependency to a package and version. We want to have more guarantees that what I'm compiling against is not manipulated and therefore include (IPFS) hashes of my dependencies.
- However, this is all still *before* considering to deploy the package to a chain. As a developer I'd want to do the Lock 1) generation efforts only once: when I've tested my stuff and I'm release version x.y.z.

As a result, the package manifest is the input to generating lock 1), i.e. the lock file can be generated based on the manifest.
Let me try and draw a scenario that I'd not find surprising if someone wrote a package that becomes very popular and gets used all over the place:

Release Manifest         ->        Release Lock        -->            Deployment Lock(s)
MyPackage-1.0.0.json -> MyPackage-1.0.0.lock --> MyPackage-1.0.0-ropsten-<timestamp>.lock
                                                                                |-> MyPackage-1.0.0-eth-<timestamp>.lock
                                                                                |-> MyPackage-1.0.0-etc-<timestamp>.lock
                                                                                |-> MyPackage-1.0.0-private-ethXYZ-<timestamp>.lock

Please notice that it's fundamental that just because a package gets installed/deployed to a particular chain it is not required to increase the version number! Even multiple deployments to the same chain can be distinguished by timestamp.
As long as the source code did not change and the integrity of dependencies from the release lock is satisfied, there is no need to proliferate the space with versions.

What happens to individual deployment lock files depends. For public chains they should definitely get published in a registry, so others can find and use them. In a lot of cases on Ethereum the author of the package might be the one who deploys his/her package and publishes the lock, but that's not a requirement.

Anyway, enough for now.
I just wanted to get my thoughts out and explain why there should be a clean separation between purely package/release-related information and chain-specific data.
One is the result of putting together a versioned package consisting of compiled binaries, ABIs, source code, manifest, and release lock.
The other is the result of taking that versioned package and installing it on a blockchain.

Feedback welcome!

pipermerriam:
@j-h-scheufen Thanks for the writeup on this.  I appreciate the time you took doing it.

Our choice for at least the initial version of this specification is to combine the two lockfiles you talk about.  This would change your diagram to something like the following.

```
Release Manifest     ->      Release Lock File
MyPackage-1.0.0.json -> MyPackage-1.0.0-ropsten-<timestamp>.lock
                                             |-> MyPackage-1.0.0-eth-<timestamp>.lock
                                             |-> MyPackage-1.0.0-etc-<timestamp>.lock
                                             |-> MyPackage-1.0.0-private-ethXYZ-<timestamp>.lock
```

Another way to think about this is that the data in your Release Lock is embedded in the Deploy Lock.

j-h-scheufen:
Thanks @pipermerriam! I'm not opposed to fusing the two lock files.
But, do I understand your last post correctly that you are indeed planning to have multiple lock files (one for each deployment to a chain) that carry the same version number? The spec suggests that simply due to the naming convention of X.Y.Z.lock this is not possible and a new version has to be created for each deployment. ... but maybe I missed latest changes?

This is really the main "beef" I have with the spec: I don't think people that want to deploy someone else's package to different blockchains (even if it's only between etc and eth) should have to contact the developer to versionize a lock file for them (or to force them to fork the original package and do it themselves). This is pure consumer-focused reasoning. And I'm not talking about private/permissioned chain clients. This applies 100% to public chains as well.

tcoulter:
@j-h-scheufen Maybe I can take a stab at this. 

> The spec suggests that simply due to the naming convention of X.Y.Z.lock this is not possible and a new version has to be created for each deployment. ... but maybe I missed latest changes?

Version numbers only matter for registries; that is, it'll be the registry's job to ensure version numbers will always be increasing, and that you can't deploy the same package with the same version twice in the same registry. 

IPFS on the other hand will accept any file. If you upload the exact same file twice, you'll receive the same hash the second time as you received the first time, meaning the file was only uploaded once. Assuming you don't put chain and address data in the lockfile as its specified currently (chain data is optional if you don't include addresses), you're guaranteed to be persisting the same file each time assuming all else is equal (i.e., the version doesn't change, etc.)

As is our goal, we're not developing a registry here. Moreover, if the registries we're assuming will exist are as (non)permissive as described above, you could still deploy two registries to the same chain without ill effect, and deploy the same package to each registry with the same version number. Given the externalities of IPFS, this will only create one file on IPFS.

> I don't think people that want to deploy someone else's package to different blockchains (even if it's only between etc and eth) ... This is pure consumer-focused reasoning. 

Arguably true, but if you're on ETC, would you want to have to run both the ETH chain and the ETC chain just to develop and deploy to ETC (that is to say you'd have to read from all your packages from the ETH chain in this case)? I can see incentive to stay inside your walled garden as that's easier.

pipermerriam:
everyone should take a look at #35

pipermerriam:
@j-h-scheufen I encourage you to take a look at #43 and specifically this use case.  https://github.com/pipermerriam/epm-spec/blob/e2cf83d8f94a5d2d1933032245f03e9e575bea28/README.md#package-which-uses-a-reusable-contract-from-a-depenency

I believe that it fits the use case that you were concerned about and should demonstrate how this use case would be covered under the current specification.

PR #35, opened 2016-12-14
Add support for defining multiple chains

pipermerriam:
### What was wrong

After some discussion, primarily sparked by #30 it seems appropriate to support deployments on multiple chains in the same release lock file.

Without this support we run into any or all of the following unfortunate or non-ideal situations.

* `owned-testnet` and `owned-mainnet` packages because some group finds it best to use the package namespace to define packages on each chain.
* Package indexes having to incorperate chain-namespaces into their package index so that people can resolve `owned[mainnet]==1.0.0`.
* Difficulty on the package manager level because you want to test on the testnet but deploy on the mainnet and in order to get the right link dependencies you either have to have a mainnet and testnet version of the same package installed or some other shenanigans.

Thanks to @VoR0220 for getting this conversation started and all of those who helped identify this as a problem.  Apologies for failing to recognize it as such when it was brought up.

### How this implements the fix

The `contracts` portion of the Release Lock File will now look something like this.

```javascript
"SimpleContract": {
  "bytecode": "0x",
  "code": "0x",
  "deployments": {
    "blockchain://<mainnet>": {
      "SimpleContract": {"address": "0x…", ...},
    },
"Token": {
  "bytecode": "0x",
  "code": "0x",
  "deployments": {
    "blockchain://<mainnet>": {
      "GoldToken": {"address": "0x…", block: "23423432", "transaction": "0xaadsfasdfas"},
      "SilverToken": {"address": "0x…"},
    },
    "blockchain://<ropsten>": {
      "GoldToken": {"address": "0x..."},
      "SilverToken": {"address": "0x..."},
    }
  }
}
```

This allows a release of a package to define deployments on multiple chains and should make the life of the package manager easier.

#### Cute animal picture

704c4283479c2f36697d880ef33830f5

### DIFF
--- a/release-lock-file-spec.md
+++ b/release-lock-file-spec.md
@@ -90,36 +90,16 @@ Sources are declared in a key/value mapping.
 * Type: Object (String: String)
 
 
-### Chain: `chain`
-
-The `chain` field defines the blockchain that should be used for any addresses
-provided with this package.  A blockchain is defined using
-a subset of the [BIP-122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki) spec, specified below:
-
-```
-blockchain://<genesis_hash>/block/<latest block hash>
-```
-
-The `<genesis hash>` represents the blockhash of the first block on the chain, and `<latest block hash>` represents the hash of the latest block that's been reliably confirmed (package managers should be free to choose their desired level of confirmations).
-
-*If* this release lock file includes any addressed contracts
-this field **must** be present. 
-
-* Key: `chain`
-* Type: String
-* Format: `blockchain://[0-9a-fA-F]{64}/block/[0-9a-fA-F]{64}`
-
-
 ### Contracts: `contracts`
 
 The `contracts` field declares information about the deployed contracts
 included within this release.
 
 * Key: `contracts`
-* Type:  Object (String: *Contract Instance Object*)
+* Type:  Object (String: *Contract* Object)
 * Format: 
     * All keys **must** be valid contract names matching the regex `[_a-zA-Z][_a-zA-Z0-9]*`.
-    * All values **must** conform to the *Contract Instance Object* definition.
+    * All values **must** conform to the *Contract* object definition.
 
 
 ### Build Dependencies: `build_dependencies`
@@ -139,26 +119,10 @@ this project depends on.
 * Type: Object (String: String)
 
 
-#### The *Contract Instance Object*
+#### The *Contract* Object
 
-A *Contract Instance Object* is an object with the following key/values.
+A *Contract* Object is an object with the following key/values.
 
-* `contract_name`:
-    * Required: Yes
-    * Type: String
-    * Format: Valid contract name matching regular expression `[_a-zA-Z][_a-zA-Z0-9]*]`
-* `address`:
-    * Required: No
-    * Type: String
-    * Format: Hex encoded ethereum address of the deployed contract.
-* `deploy_transaction`:
-    * Required: No
-    * Type: String
-    * Format: [BIP122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki) URI which defines the transaction in which this contract was created.
-* `deploy_block`:
-    * Required: No
-    * Type: String
-    * Format: [BIP122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki) URI which defines the block in which this contract was created.
 * `bytecode`:
     * Required: No
     * Type: String
@@ -199,6 +163,35 @@ A *Contract Instance Object* is an object with the following key/values.
                     * Required: No
                     * Type: Integer
                     * Format: Positive Integer
+* `deployments`
+    * Required: No
+    * Type: Object(String: Object(String: *Deployment* object)
+    * Format: 
+        * All keys **must** be valid BIP122 URIs matching the regular expression `blockchain://[0-9a-fA-F]{64}/block/[0-9a-fA-F]{64}`
+        * All values **must** be objects which conform to the following format
+            * Type: Object(String: *Deployment* object)
+            * Format:
+                * All keys **must** be strings conforming to the regular expression `[_a-zA-Z][_a-zA-Z0-9]*`.
+                    * In the case that there is a single deployed instance this **should** use the same name as the contract.  
+                    * In the case that there are multiple deployed instances of the same contract a unique name should be used for each deployed instance.
+                * All values **must** be objects conforming to the *Deployment* object format (see below).
+
+
+#### The *Deployments* Object
+
+
+* `address`:
+    * Required: Yes
+    * Type: String
+    * Format: Hex encoded ethereum address of the deployed contract.
+* `deploy_transaction`:
+    * Required: No
+    * Type: String
+    * Format: [BIP122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki) URI which defines the transaction in which this contract was created.
+* `deploy_block`:
+    * Required: No
+    * Type: String
+    * Format: [BIP122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki) URI which defines the block in which this contract was created.
 * `link_dependencies`:
     * Reqired: Required to have an entry for each link reference found in either the `bytecode` or `runtime_bytecode` fields.
     * Type: Object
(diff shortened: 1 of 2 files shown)


VoR0220:
Again, the problem is that this makes the lock file constantly updating...which it shouldn't be. I again propose that we segregate blockchain deployment properties from code properties. Am happy though that you see this as an issue now.

pipermerriam:
Some thoughts on why I like this approach over a separate file as was proposed in #30 

* Separate chain file adds more files that the package index needs to track.
* Chain file would be required to reference the manifest file which means the manifest file cannot *know* about all of the chain files.  This effectively adds mutability into releases since the chain file(s) would have to be generated after the release was generated.  I much prefer that the Release Lock File is self contained and that once you have that file, you have the whole of the release.
* I think it's a bad idea to allow *tacking* on new addresses to an existing release, but I do see the importance of being able to define where the release lives on different chains.  This satisfies both.

pipermerriam:
> Again, the problem is that this makes the lock file constantly updating...which it shouldn't be.

I'm not proposing that Release Lock Files be modified.  In fact, I am adamantly opposed to this.  What this does is require that you define *all* deployed locations at the time that the release is generated.

*If* you need to deploy to a new chain after a release has been generated then you would do so with a new release, likely bumping the `patch` portion of the version and include the new chain definition as well as all of the old ones.

VoR0220:
One is not a release, one is a deployment. There is a major difference between the two. But it's fine. I don't think I'm going to get my way here, just happy that we've had some small impact. This gets my blessing.

pipermerriam:
And some more thoughts on how I see the developer process interacting with this.

1. I write my code and test against ephemeral chains (no releases)
2. I get to a point where I'm ready for the testnet but not mainnet.  Release as `1.0.0-alpha1` which only includes testnet addresses.
3. When it's *ready* release as `1.0.0` which includes both testnet and mainnet addresses.
4. Later *morden* gets canned and this new hipster chain *ropsten* shows up.  Release as `1.0.1` which only differs from `1.0.0` in that it includes an additional chain definition.

j-h-scheufen:
@pipermerriam Thanks for keeping this topic alive!
> require that you define all deployed locations at the time that the release is generated
This implies that the author of a package is also always his own customer, i.e. in control of all deployments of a package. If that was the case, then you are correct, the author can just bump the patch level to include more deployment locations if needed ...
But what about the example with a "BankInABox" package I described in #30? If a package author simply *publishes* a package and other people later deploy it as a dependency in their solutions. Now the consumers are forced to deal with generating new release locks, is that not correct?

I think it can be confusing to speak of a "release" as being both: publishing and deployment. This is very unique to the blockchain world, but normally a "release" is just the act of publishing something for consumption by others. There are use cases where (as an author) I'd want to just publish a "release" without necessarily deploying it to a particular blockchain.

Again, you are 100% spot on if you assume that the act of "releasing" always includes a deployment on one or more chains, but I don't think we should enforce this restriction on developers.

tcoulter:
> other people later deploy it as a dependency
As far as I understand, those deployments aren't part of the package. In my understanding, the package isn't meant to maintain all deployments of a specific set of code, but only specific (we'll call them "canonical") deployments. i.e., a `library` could and should have a canonical deployment. A token meant to represent a specific asset should have a canonical deployment. `owned`, a package mainly meant to be used in inheritance, likely should't have a canonical deployment. Does that make sense?

>  I'd want to just publish a "release" without necessarily deploying it to a particular blockchain.
This is entirely possible within the current spec.

> Again, you are 100% spot on if you assume that the act of "releasing" always includes a deployment on one or more chains, but I don't think we should enforce this restriction on developers.
We don't actually, if I understand you correctly. You're not required to have deployments as part of a package.

pipermerriam:
> If a package author simply publishes a package and other people later deploy it as a dependency in their solutions.

This is the part of that use case that I do not fully comprehend as it seems wrong.  If someone else deploys my package, while the code and everything may be the same, it is **not** my package and I am **not** ok with their deployment being listed under *my* package.  My reasoning is that this opens up significant security holes since they could easily deploy the package with some sort of malicious configuration and if it is listed under my package name then they get to piggy back on my credibility.

What am I missing.

tcoulter:
Ah, perhaps it wasn't clear in @pipermerriam's PR. `deployments` is an optional key. Whens specified, it looks like the example above.

j-h-scheufen:
@tcoulter Correct. The spec does not require you to have deployments as part of the release.
@pipermerriam I understand your point. I'm making a distinction between "authorship" and "use". I as an author vouch for my code and my release version. And I would also vouch for the deployments that I did myself. Someone who uses the package with a malicious configuration should get differentiated from my "honest" deployment. But now we're back to the point why there should be two lock file types, one for the act of publishing a release version and one for deploying a release version.

pipermerriam:
So maybe I now understand it.

*BankInABox* is a re-usable set of contracts that do not come with pre-deployed addresses.  Every person who *uses* this package deploys their own versions of these contracts.

So this would work as follows:

* The author of the *BankInABox* package would publish and release their package.  This Release Lock File would not include any deployment information because the package is intended to be used as base contracts for creating your own banking service.
* Suzy decides to launch a new banking service and uses *BankInABox*.  She installs *BankInABox* as a package dependency, and then deploys her versions of the contracts.  When she generates a release, she **does** include information about the deployed instances, but they are *her* contracts and thus the addresses would be published under *her* package, **not** the *BankInABox* package.

Does this match what you are thinking about?

j-h-scheufen:
Yes, that matches the use cases I'm thinking about.
Question though: how would Suzy declare and "lock" the guarantee that the Bank contract address she has in her deployment is based on your published release "BankInABox-1.2.4"?
She has a unique deployment lock due to her having her own Bank contract instance, right? But to get there she needs to fork the source code of your package and build/deploy it herself to generate the required unique lock file that is detached from your lock file (according to the current spec). Isn't this unnecessary and overkill? The source code hasn't changed! Suzy should just refer to your published release lock and still be able to create her specific deployment lock.
Please correct me if the above description is not what the current spec prescribes.

tcoulter:
> She has a unique deployment lock due to her having her own Bank contract instance, right?

If she's publishing her package as its own package, yes. If she's not, this lockfile won't be published to IPFS or a registry and will simply exist in a source code repository (think `npm`, `package.json` and associated lockfiles for projects that aren't published as packages).

> But to get there she needs to fork the source code of your package and build/deploy it herself to generate the required unique lock file that is detached from your lock file (according to the current spec). 

If she's not publishing her own package, there's no forking; it'd instead simply be using `BankInABox` as a dependency.

If she is publishing her own package, she can include the deployed contract objects in the lockfile without including any other information or source code about those contracts, since her package would link to `BankInABox` as a dependency.

> Isn't this unnecessary and overkill? 

Given the above I don't see any overkill in that if she's creating a separate package, a lockfile is required. No extra source code needs to be included and even if it was it wouldn't add new files onto the network (see below). 

> The source code hasn't changed! 

Correct. And as I understand it it wouldn't need be redeployed either since her lockfile references it as a dependency. Even if it was, it'd be given the same IPFS hash meaning it'd reference the same file on the network, meaning it'd only uploaded once, ever.

> Suzy should just refer to your published release lock and still be able to create her specific deployment lock.

This is exactly how it's set up within the [spec](https://github.com/ethpm/epm-spec/blob/master/release-lock-file-spec.md), in that the lockfile references dependencies. She'd do so as a new package, however.

pipermerriam:
@j-h-scheufen 

> But to get there she needs to fork the source code of your package and build/deploy it herself to generate the required unique lock file that is detached from your lock file

No.  She would install `bank-in-a-box==1.2.4` which would make those contracts available in her local project.  When she generates a release she will include the deployed addresses of whatever contracts she used from `bank-in-a-box` in her release lockfile.  She doesn't need to vendor or fork the source code because her release will point at the `bank-in-a-box==1.2.4` release lockfile which contains the source code for those contracts.  

I believe this is your desired scenario, for her to be able to make use of contracts from a different package without needing to copy those contracts into her package and to still maintain the ability to verify all of the deployed instance bytecode.

Correct me if I've gotten something wrong.

pipermerriam:
Are there objections to merging this.  I will plan on implementing the use case brought up by @mhhf separately.

j-h-scheufen:
Thanks, @tcoulter!
I'm afraid I'm still not seeing Suzy's scenario being supported the way it should be. The terms "publish", "release", "deploy" are being conflated IMO when in blockchain land there should be a clear distinction between the publishing of a versioned package to a repository and the deployment of it to a chain.
Yes, Suzy can use it (as-is) as a dependency when she develops her solution, but if she wants to deploy her solution to a chain that never had the Bank ABI registered, she'd still have to create her own (custom version) release lock for BankInABox (again: because this spec does not differentiate between "publishing a release and deploying a release). This puts too many hurdles in the way of developers and stifles the fostering of a community by limiting the reusability of artifacts.

@pipermerriam This PR is a step in the right direction nevertheless. So, I support it and you should go ahead! I'm going to try and write down a more detailed "Suzy" scenario to verify (or falsify) my concerns ... after all, I'm not ruling out that I'm mistaken here, but I hope you guys can see that I care to get this right.

VoR0220:
So I think I'm better understanding both points of view right now, and I can see a difference in the philosophies better now. I think that a third file might not be needed, but depending on how the registry is structured (ie, are you making this registry on chain with the EVM), then I think it might be possible to simply reference a bip address of the contract with the state of deployments at this point and time. This actually could operate within the spec.

pipermerriam:
@j-h-scheufen I am inclined to think that this is a disconnect in understanding because I'm pretty sure your case works.  I began work writing up detailed examples of the range of use cases that we have thought about in #39 and will let you know when there's something there for you to peek at.

PR #36, opened 2016-12-15
update spec to only allow IPFS URI's in the build dependencies

pipermerriam:
### What was wrong

In the Release Lock File the `build_dependencies` allow declaring the dependency as

```javascript
{
  'build_dependencies': {
    "owned": "1.0.0",
  }
}
```

This is problematic because it means that installation is no longer deterministic because it depends on the registry.

### How was it fixed

Remove the allowance of specifying version number so that all `build_dependencies` must be IPFS URIs

#### Cute Animal Picture

grizz5

### DIFF
--- a/release-lock-file-spec.md
+++ b/release-lock-file-spec.md
@@ -108,15 +108,12 @@ included within this release.
 the `dependencies` field defines a key/value mapping of ethereum packages that
 this project depends on.
 
-* All keys **must** be valid package names matching the regular expression `[a-z][-a-z0-9]{0,213}`
-* All values **must** conform to *one of* the following formats:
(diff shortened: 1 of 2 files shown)


VoR0220:
For what it's worth, I would edit this to allow for IPFS URI and traditional URI (a simple github link would suffice).

mhhf:
github uri introduces uncertainty. We are using IPFS in the first place as checksum hash and only in the second place as location pointer. Location pointing should be solved on a higher level. E.g. you can build a registry which takes a hashsum and returns a Github URI.

VoR0220:
I suppose I can mesh with that thinking.

PR #37, opened 2016-12-15
Refine how link dependencies are defined.

pipermerriam:
### What was wrong

The previous way to declare link dependencies sucked

### How was it fixed

Came up with something better.  Will fill this in better later.  I need to wrap up my day.

#### Cute animal picture

muppets3


pipermerriam:
Closing in favor of https://github.com/ethpm/epm-spec/pull/41

PR #39, opened 2016-12-15
begin writing up use case narative

pipermerriam:
This is a work in progress.


pipermerriam:
redirecting at a different branch.

Issue #40, opened 2016-12-15
Alternative structure and behavior for contract definitions


tcoulter:
## Overview

I think #38 doesn't provide much utility and may contribute many edge cases. Similarly, I think the `contracts` keyword is getting overloaded and needs refactoring. 

This issue represents a response to #38 and more recent PRs, and is an attempt to open a discussion about restructuring some of the lockfile's data.

## Background

It was hard for me to reason about this, so I want to start by introducing some definitions that help describe my thinking.

* `contract`: A `contract` is source code that describes an object that can be deployed to a blockchain.
* `instance`: An `instance` is a deployed version of that contract with an address on a specific chain.
* `type`: All instances have a `type`; a `type` is the specific `contract` used during deployment to create that specific `instance`. For simplicity, let's assume that if an instance is said to have type `A`, then the source code used to deploy it was defined by a contract called A, using the code `contract A {}`.
* `contract artifacts`: Results of compilation and deployment, including abi, binary, deployed address, etc.

Given the structure proposed in #38, using all the names available to us would look like the JSON blob below. I've defined my understanding of these names and given each name an identifier. 
 
* `ContractName|Alias`: "contract name or alias". When `ContractName`, this will equal the type name. When `Alias`, this will equal an arbitrary name we want to call this contract.
* `TypeName`: "type name", same as `ContractName`. Not arbitrary (i.e., not `Alias` or `InstanceName`).
* `InstanceName`: "instance name", an arbitrary name we can give to a specific instance.

```javascript
{
  // ...
  contracts: {
    "ContractName|Alias": {
      // ...
      "contract_name": "TypeName"
      "deployments": {
        "blockchain://...": {
           "InstanceName": {...}
        }
      }
    }    
  }
}
```

Some assumptions I'm making:

* `TypeName` is intended to be used to reference a specific contract within the process of code verification.
* A dependency may declare a contract that has the same `TypeName` as a contract within the root package, or two dependencies may declare contracts with the same `TypeName`. 
* `InstanceNames` are required to be unique per contract/alias per blockchain, but will not necessarily be unique across contract objects.
* The `contracts` item in the lockfile exists for two (and only two) reasons:
  * To describe contracts that are deployed as part of this package, including their associated artifacts.
  * To describe contract types and their associated artifacts when the package does not include source code.

Given this structure and my assumptions, here's some observations:

[...]

PR #41, opened 2016-12-15
Lots and lots of things

pipermerriam:
What does this do?

* Adds RFC2119 definition for the various RFC style keywords like MUST and SHOULD
* Convert to the naming scheme "lockfile" instead of "lock file"
* Define a number of other terms which are used throughout the spec document.
* Setup terminology for talking about contracts, their *types*, and their deployed instances.
* Modified convention for the filename to match what is going on in populus and truffle.
* Fix the `lockfile_version` type.
* Removed `package_manifest` field
* Added new fields so that all data that is in the Package Manifest can be included in the Release Lockfile.
* Removed the *Contract Instance Object* concept.
* Added the `contract_types` and `deployments` which better represents the data that was previously in the *Contract Instance Object* (See #40 )
* Resolved `build_dependencies` key discrepancy.
* Changed the format that `link_dependencies` are declared.

This is a *bunch* of stuff.  Anything controversial can get split out into it's own separate PR.  I know everyone is busy but please review as soon as possible.

#### Obligatory Cute Animal Picture

baby-cute-puppy-corgi


VoR0220:
your cute animal has not rendered. I'll look at this in a bit as it's a lot to digest and I have other things that need attending to.

release-lock-file-spec.md
```diff
@@ -23,52 +152,69 @@ single JSON object.
...
-When creating a release, convention is to name this document `<version>.json`.
+When creating a release, convention is to name this document using one of the following conventions.
 
-* The `1.0.0` Release would be named `1.0.0.json`
+* In cases where multiple *Release Lockfiles* are being store alongside one another, convention is to use `<version>.json`.  For example, the `1.0.0` release would be named `1.0.0.json`.  
+* In cases where a single *Release Lockfile* is being stored, convention is to use `lock.json`.
 
-Package managers may keep their own copy of this file as part of their package installation process.  In these cases it is up to the package manager to decide on a naming convention.  Package managers **should** keep the `.json` file extension.
+Package managers **should not** rely on these naming conventions nor are they
+required to follow them.
```
tcoulter: Not to be a curmudgeon, but if package managers shouldn't rely on them, should this section exist in the spec? On my end, I'm actually calling the files `lock.json` for consistency of implementation. This is a package manager implementation decision, so I think we should remove it from the sepc.

pipermerriam: So all of these statements use the **should** keywords which are meant to say, if you don't use this pattern, we really recommend that you have a pretty good reason.  

Including these types of things helps with general consistency across implementation in areas like this where we don't **need** compatibility but that it's nice from the user perspective for things to at least be familiar when they switch between tools.

release-lock-file-spec.md
```diff
@@ -1,18 +1,147 @@
...
+`[-a-zA-Z0-9_]`.
+
+Package managers **should** use the unprefixed hexidecimal representation of
+the Keccak (SHA-3) hash of the *contract type* bytecode in it's binary
+representation as the `identifier` value.
+
+In the case that the *contract types* with name collisions have the same
+bytecode such as two abstract contracts, package managers **should** fall back
+to the Keccak (SHA-3) hash of the source file in which each *contract type* is
+defined.
```
pipermerriam: I think I see your point on the *emptyness* of the collision case.

So if I understand correctly, you would be happy with this if there were one naming scheme: `ContractName([keccak(bytecode)])`?

release-lock-file-spec.md
tcoulter: Perhaps. But to use an example, digging inside `node_modules` or python eggs (or whatever they're called) is not something a user would do on a regular basis. The only time I've dug into package locations disk is when I was publishing my own package (in the case of python; I'd install the package globally and then see what got installed because it was unclear a priori); or I was dealing with some weird `npm`'ism in how `npm` breaks import statements in very rare cases.

This is all to say the innards of the `installed_contracts` directory isn't something that's user facing. If `yarn` and `npm` want to do it differently, for instance, that's up to them (yarn saves a `yarn.lock`; npm does not). I just don't think the spec needs to worry about this aspect of the implementation.

pipermerriam: Removing this section from the spec.

release-lock-file-spec.md
tcoulter: Correct. 

It might be useful to mention that abstract contracts shouldn't be placed in the `contract_types` object, just to remove all possibility of conflicts.[1]

------------

1. The other possibility of conflict that I've thought of is the case where two files have the exact same bytecode and the exact same contract name. This is in effect a direct copy of a source file, but perhaps places in a different directory within the source. This would be bad coding practice, generally, but shouldn't be seen as an actual conflict as those contracts would be verified as the same contract.

release-lock-file-spec.md
```diff
@@ -1,74 +1,196 @@
...
+### Custom
+
+#### Prefixed vs Unprefixed
+
+A prefixed hexidecimal value begins with `'0x'`.  Unprefixed values have no
+prefix.  Unless otherwise specified, all hexidecimal values should be
+represented with the `'0x'` prefix.
+
+* Prefixed: `0xdeadbeef`
+* Unprefixed: `deadbeef`
```
VoR0220: Big ++ from me on this.

release-lock-file-spec.md
```diff
@@ -1,74 +1,196 @@
...
+#### Contract Alias
+
+This is a name used to reference a specific *contract type*.  Contract
+aliases **must** be unique within a single release lockfile.  
+
+The *contract alias* **must** use the naming scheme
+`<contract-name>[<bytecode-identifier>]` where `<contract-name>` is the
+*contract name* and `<bytecode-identifier>` is the unprefixed hexidecimal
+representation of the Keccak (SHA-3) hash of the *contract type* bytecode in
+it's binary representation
```
VoR0220: What was the reason for adding this again? It's a bit confusing to me. I know @mhhf said something about this.

release-lock-file-spec.md
```diff
@@ -1,74 +1,196 @@
...
+conform to the regular expression `[a-zA-Z][a-zA-Z0-9_]*`.
+
+In cases where there is a single deployed instance of a given *contract type*
+package managers **should** use the *contract alias* for that *contract type*
+for this name.
+
+In cases where there are multiple deployed instances of a given *contract type*
+package managers **should** use a name which provides some added semantic
+information as to help differentiate the two deployed instances in a meaningful
+way.
```
VoR0220: I'm for this. ++

release-lock-file-spec.md
```diff
@@ -90,117 +212,377 @@ Sources are declared in a key/value mapping.
...
 
-### Contracts: `contracts`
+### Contract Types: `contract_types`
 
-The `contracts` field declares information about the deployed contracts
-included within this release.
+The `contract_types` field holds information about *contract types* which
+have been included in this release lockfile.  Release lockfiles **may not**
+include any *contract types* which are not present in the project's source
+code.
```
VoR0220: Could do with a `.sol` example here. I assume you mean `contract C` would mean that the  type is C but from initial reading it's unclear.

release-lock-file-spec.md
```diff
@@ -90,117 +212,377 @@ Sources are declared in a key/value mapping.
...
-        * All keys **must** be strings which are formatted as valid link targets.
-        * All values **must** conform to *one of* the following formats:
-            * A hex encoded ethereum address.
-            * A [json pointer](https://tools.ietf.org/html/rfc6901) to another *Deployment* object in the release lock file.  The `address` from the resolved *Deployment* object should be used.
-            * An IPFS URI with a JSON point in the fragment portion of the URI.  The IPFS hash must resolves to a valid release lock file.  The json pointer **must** resolves to a *Deployment* object within the release lock file.  The `address` from the resolved *Deployment* object should be used.
-
-
-#### BIP122 URIs
+    * Keys **must** be valid package names matching the regular expression `[a-z][-a-z0-9]{0,213}`
+    * Values **must** be valid IPFS URIs which resolve to a valid *Release Lock File*
```
VoR0220: I talked this over with my team and they want to abstract this so that it could include other hashing schemes, but for now this will do...just something to keep in mind.

release-lock-file-spec.md
```diff
@@ -90,117 +212,377 @@ Sources are declared in a key/value mapping.
...
+
+Definitions for different objects used within the release lockfile.  All
+objects allow custom fields to be included.  Custom fields **should** be
+prefixed with `x-` to prevent name collisions with future versions of the
+specification.
+
+
+### The *Package Meta* Object
+
+The *Package Meta* object is defined to have the following key/value pairs.
```
VoR0220: ++ on putting authors and license into the meta object...this is a good separation.

release-lock-file-spec.md
```diff
@@ -1,74 +1,196 @@
...
 
-* The `1.0.0` Release would be named `1.0.0.json`
+The act of replacing *link references* within some bytecode with *link values*.
 
-Package managers may keep their own copy of this file as part of their package installation process.  In these cases it is up to the package manager to decide on a naming convention.  Package managers **should** keep the `.json` file extension.
+
+#### Contract Type
+
+This term is used to refer to a specific contract or library in the package
+source.
```
pipermerriam: What do you think about adding the sentence *"This term can be used to refer to an abstract contract, a normal contract, or a library."*

release-lock-file-spec.md
pipermerriam: I rewrote this to hopefully be less cryptic.  Will you let me know if the revised version makes more sense.

release-lock-file-spec.md
tcoulter: Referring a contract as a `library`, `abstract contract`, `interface`, etc. is an unfortunate overloading of the word "type". Both usages are correct here, but they describe different things. For the purposes of this document we want to describe the _name_ of the object that can be placed on the blockchain, and not necessarily the _type_ of object that is (i.e., `contract`, `abstract contract` or `library`). Some examples:

With the following code, the type is `A`:

```
contract A {
  public uint value;
  function A() { 
    value = 2;
  }
}
```

This code also has a type `A`: 

```
library A {
  function add(uint a, uint b) returns (uint) { return a + b; }
}
```

And so does this code: 

```
contract A {
  function add() returns (uint);
}
```

So it's not the _type_ of object, but rather than the _name_ of the object. We're using the word "type" here because instances have a type, in that we can say an instance of contract `A` is of type `A`.

VoR0220: @pipermerriam if we are following @tcoulter 's example here, then we should display an example of it. Do that and I think it's good to go.

release-lock-file-spec.md
```diff
@@ -1,74 +1,197 @@
...
 
-Package managers may keep their own copy of this file as part of their package installation process.  In these cases it is up to the package manager to decide on a naming convention.  Package managers **should** keep the `.json` file extension.
+The act of replacing *link references* within some bytecode with *link values*.
+
+
+#### Contract Type
+
+This term is used to refer to a specific contract in the package source. This
+term can be used to refer to an abstract contract, a normal contract, or a
+library.
```
tcoulter: Maybe I'm getting pedantic, but it should be _the name of_ the abstract contract, normal contract, or library, correct?

I want to make sure the set of `contract types` != `["abstract", "contract", "library"]`, and instead represent (e.g.) `["Token", "MyRegistry"]`, etc.

pipermerriam:
Will get an example added.

release-lock-file-spec.md
mhhf: Maybe you should specify the exact Keccak here: I suppose you mean 256-bit?
However I'd argue to use 128-bit or even 64-bit since collisions aren't really a problem here and we just use it for distinction and 64-bit is a more readable:
```
Token[e843d7a97cf7c772bd9d2880878920c33f06afee65cc2e87b9d19e3416d61249]
vs.
Token[e843d7a97cf7c772]
```

release-lock-file-spec.md
mhhf: Ipfs uses [multihash](https://github.com/multiformats/multihash) which supports exactly this: all common hash algorithms as well as space for new ones.

release-lock-file-spec.md
```diff
@@ -90,117 +224,374 @@ Sources are declared in a key/value mapping.
...
+### Keywords: `keywords`
+
+The `keywords` field provides relevant keywords related to this package.  
+
+* Required: No
+* Key: `keywords`
+* Type: List of Strings
+
+
+### Links: `links`
```
mhhf: Not sure I agree with putting ephemeral information in the lockfile. I'd argue that lock-files should only contain deterministic information and links should be solved on a meta level (e.g. registry).

release-lock-file-spec.md
VoR0220: @mhhf lets stick to the 256-bit...far less confusing and more standardized. You could always take the first N bytes of the hash anyways, couldn't you? Sort of how the ABI does it?

release-lock-file-spec.md
VoR0220: it does? Interesting. Thanks for the tip @mhhf

release-lock-file-spec.md
```diff
@@ -1,74 +1,208 @@
...
+
+A link value is the value which can be inserted in place of a link reference.
+
+
+#### Linking
+
+The act of replacing *link references* within some bytecode with *link values*.
+
+
+#### Contract Type
```
VoR0220: bit of a nitpick, eh?

release-lock-file-spec.md
VoR0220: isn't that why this is in the meta object?

mhhf: but isn't the meta object referenced from the lockfile?
why not referencing the lockfile from the meta file and solve the meta file on a higher level where we introduce nondeterminism?

pipermerriam: @mhhf the meta object is embedded within the lockfile and none of the information is required.  This change makes lockfiles fully self-contained as they no longer reference the manifest.

release-lock-file-spec.md
```diff
@@ -238,6 +242,10 @@ which can be found in the source files for this package.  Release lockfiles
     * Values **must** conform to the *Contract Type* object definition.
 
 
+Packages **should** not include abstract contracts in the *contract types*
```
nmushegian: bold **should not**

nmushegian: Or is it meant to say "**should** include"

pipermerriam: fixed

pipermerriam: The current language is correct.  

We encourage leaving abstract contracts out of this section because:

* They allow naming collisions to occur
* inclusion of them is not useful in a release lockfile.  (Logic is as follows)

Inclusion of abstract contracts is not useful in release lockfiles because:

* If you only want to provide the interface information, you should just do so using the *real* contract and not the interface.
* Contract Types are only referenced internally from the *contract instance* which means that it is impossible for a contract instance to reference an abstract contract since abstract contracts are not deployable.

Neither of these scenarios makes inclusion of an abstract contract the correct choice when the actual contract is available.  The spec does allow inclusion of an abstract contract here but it's rarely going to be the right choice.

PR #43, opened 2016-12-16
Documentation and explanations of the use cases we are covering and how they are implemented.

pipermerriam:
Work in progress.

- [x] stand-alone base contract
- [x] base contract with dependency
- [x] reusable contract
- [x] contract with local deployed instance
- [x] contract with deployed instance from a dependency
- [x] contract with deployed instance with local link dependency
- [ ] contract with deployed instance with link dependency from package dependency
- [ ] multiple deployments of the same contract
- [ ] deployments on multiple chains


nmushegian:
> contract with library dependency and deployed instance

Can you make a usage example of a non-lib instance dependency, e.g. [weth](https://github.com/nexusdev/weth)? I see two basic ways this is handled and want to understand what the ethpm devs imagine themselves doing. One is do the linking the constructor args level which means the deploy tool handles it, which I think is out of scope of the spec so far. The other is to wrap things like weth into libraries with a hard-coded address so that they can be shipped (equivalent once address templating is more general).

pipermerriam:
@nmushegian This spec doesn't differentiate between a library and a contract.

For the case you are talking about all of the required information is available to you in the Release Lockfiles and it will be up to the package manager to pull it out and use it.

PR #51, opened 2016-12-29
update json schema

pipermerriam:
What this does.

* Fills in the json schema for the release lockfile
* Fixes two errors in the spec
    - The `transaction` field was not intended to be a BIP122 URI.  Since it is already nested under a block URI within the dependency this is redundant.
    - The `block` field was not intended to be a BIP122 URI.  Since it is already nested under a block URI within the dependency this is redundant.
* Updates the examples to be fully compliant with the spec.
* Adds a simple script that will validate all of the example lockfiles against the spec.


examples/transferable/1.0.0.json
```diff
@@ -0,0 +1,21 @@
+{
+  "lockfile_version": "1",
+  "version": "1.0.0",
+  "package_name": "transferable",
+  "meta": {
+    "license": "MIT",
+    "authors": [
+      "Piper Merriam <pipermerriam@gmail.com>"
+    ],
+    "description": "Reusable contracts which implement a priviledged 'owner' model for authorization with functionality for transfering ownership.",
```
VoR0220: *privileged

tcoulter:
FYI: I'm generally :+1: on these until I start using them. That should be soon, but I'll submit any problems I have (if any) as separate issues/PRs. Thanks @pipermerriam!

PR #53, opened 2017-01-28
Piper/fix inconsistent key name for package meta

pipermerriam:
The spec was inconsistent with the key it used for the Package Meta.

I chose to go with what was present in the JSON schema because it's precise (and I also know that @tcoulter and my implementations are already using that key.


pipermerriam:
@mhhf @iurimatias @VoR0220 any objections on this fix?

VoR0220:
it gets the go ahead from me. Merge away.

PR #54, opened 2017-02-02
fixes ipfs uri to generalized format

mhhf:


pipermerriam:
Given this is backwards compatible this feels fine to merge and backfill into the v1 spec.  Any objections @iurimatias @tcoulter @VoR0220

VoR0220:
The only objection I would have is to give a detailed reasoning for the regex expression as the way it is. Other than that I'm good.

tcoulter:
I don't have an intuitive understanding of other URLs. If there's a URL that's different than `ipfs://<multihash>`, I may not currently process it correctly as I rely on that multihash specifically. I would love a detailed reasoning as well.

pipermerriam:
@tcoulter reasoning comes from this post from the go-IPFS repo: https://github.com/ipfs/go-ipfs/issues/1678 and specifically juan's comment here https://github.com/ipfs/go-ipfs/issues/1678#issuecomment-139842752

VoR0220:
...okay...then let's put that in the spec.

pipermerriam:
@VoR0220 this PR gets the written version of the spec inline with what is already present in the jsonschema version of the spec.

The result of this will be that both the written version and the jsonschema version will allow all of

- `ipfs://Qm...`
- `ipfs:/Qm...` (previously disallowed in the written version)
- `ipfs:Qm...` (previously disallowed in the written version)
- `ipfs://Qm.../some/path`
- `ipfs:/Qm.../some/path` (previously disallowed in the written version)
- `ipfs:Qm.../some/path` (previously disallowed in the written version)

So 👍  or 👎

pipermerriam:
@VoR0220 I think I mis-understood your comment.  Are you suggesting we add that link to the specification to indicate where the URI format comes from?

VoR0220:
Not add the link because that can disappear, moreso describe in plaintext where the URI format comes from.

pipermerriam:
+1 Dennis? Think you can get that in and we can get this merged.

mhhf:
added a short explanation, hope its enough.

Should we also shorten the regex to `^ipfs:.*$`? since its technically the same thing as saying `^ipfs:/?/?.*$` ?

PR #58, opened 2017-02-21
Piper/address deficiency in link dependencies

### LINKED ISSUE #57: Deficiency in how link dependencies are specified.
Consider the case where we have three packages `A`, `B`, and `C` such that:

- `A` depends on package `B`
- `B` depends on package `C`
- `C` contains a library with a deployed instance of that library.
- `B` contains a re-usable contract which uses the library from package `C`
- `A` contains a deployed instance of the re-usable contract from package `B`

According to the spec link dependencies must either reference a deployed contract instance from within the same package, or from one of the dependencies which is declared in the `build_dependencies`.  In this case, the address for the link dependency is actually sourced from a dependency of a dependency.

Source: https://github.com/ethpm/ethpm-spec/blob/master/release-lockfile.spec.md#value-value

In order to comply with the spec, `C` would have to be included as a dependency of `A` which is far from ideal.

pipermerriam:
Fixes #57 

Allow pointing arbitrarily deep into the package dependency tree for link reference values.


pipermerriam:
First stab at trying to define this.  Would love any assistance on how to make this a bit more understandable.  I'm planning to add examples.

pipermerriam:
Example has been added.

Personally, it is *easier* to backfill this into the v1 spec since this changed doesn't invalidate any existing lockfiles, only expands the allowed format for lockfiles.  The alternative is to support two versions of the spec which I'm personally not keen on.  I only want to do this if we have agreement from everyone though.  Who's got thoughts on this?

@mhhf @VoR0220 ?

PR #59, opened 2017-03-21
Remove top-level additionalProperties restriction

gnidan:
To allow `x-*` options, et al.

https://github.com/ethpm/ethpm-spec/blob/master/release-lockfile.spec.md#document-specification
> The following fields are defined for the release lockfile. Custom fields may be included. Custom fields should be prefixed with x- to prevent name collisions with future versions of the specification.

Raised in
https://gitter.im/ethpm/Lobby?at=58d1b1bd6d7eb18404f0fc01

### DIFF
--- a/spec/release-lockfile.spec.json
+++ b/spec/release-lockfile.spec.json
@@ -7,7 +7,6 @@
     "version"
   ],
   "version": "1",
-  "additionalProperties": false,
   "properties": {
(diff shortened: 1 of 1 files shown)


gnidan:
Tested locally from within `ethpm-js`, schema allows additional options correctly.

iurimatias:
:+1:

PR #80, opened 2017-12-15
add extra spec about serialization format for json document

### LINKED ISSUE #77: Define JSON strict serialization format so that lockfiles with the *same* content always have the same hash.
Currently the spec allows for any serialization of the JSON.

* tightly packed: `{"key-a":"value-a","key-b":2}`
* loosely packed: `{"key-a": "value-a", "key-b": 2}`
* or even multi-line:

```javascript
{
  "key-a": "value-a",
  "key-b": 2
}
```

While all three of these have identical contents, they would not hash to the same value.

I propose we specify that lockfiles **must** be serialized in the *tightly packed* format above meaning no extra whitespace or newlines.

pipermerriam:
Fixes #77 

### What was wrong

Two lockfiles with identical content can easily end up hashing to a different value due to differences in whitespace used when serializing the JSON.

### How was it fixed

Define a *strict* definition of serialization which is tightly packed with sorted keys and utf8 encoded.

### DIFF
--- a/release-lockfile.spec.md
+++ b/release-lockfile.spec.md
@@ -150,8 +150,12 @@ way.
 
 ## Format
 
-The canonical format for the release lockfile is a JSON document containing a
-single JSON object.  
+The canonical format for the release lockfile JSON document containing a
+single JSON object.  Lockfiles **must** conform to the following serialization rules.
+
+* The document must be tighly packed, meaning no linebreaks or extra whitespace.
+* The *all* keys in all objects must be sorted alphabetically.
+* The document **must** use utf8 encoding.
 
 
 ## Document Specification


release-lockfile.spec.md
```diff
@@ -150,8 +150,12 @@ way.
 
 ## Format
 
-The canonical format for the release lockfile is a JSON document containing a
-single JSON object.  
+The canonical format for the release lockfile JSON document containing a
+single JSON object.  Lockfiles **must** conform to the following serialization rules.
+
+* The document must be tighly packed, meaning no linebreaks or extra whitespace.
+* The *all* keys in all objects must be sorted alphabetically.
```
pipermerriam: We should probably also add that lockfiles with duplicate keys are invalid.

release-lockfile.spec.md
```diff
@@ -150,8 +150,12 @@ way.
 
 ## Format
 
-The canonical format for the release lockfile is a JSON document containing a
-single JSON object.  
+The canonical format for the release lockfile JSON document containing a
+single JSON object.  Lockfiles **must** conform to the following serialization rules.
+
+* The document must be tighly packed, meaning no linebreaks or extra whitespace.
```
carver: "must" is the spec usage here, so should probably be **must**.
Why not just "all" whitespace instead of "extra"?
Also, I am guessing this means no newline at the end of the file, but better to be explicit.

release-lockfile.spec.md
carver: What are "the *all* keys"?

PR #86, opened 2018-05-03
Rename lockfile to package

### LINKED ISSUE #73: Rename lockfile to package
I suggest we rename lockfile to package as it's a friendlier term and less confusing.

njgheorghita:
Use of "Lockfile" can be confusing and misleading as Package Managers will be expected to implement their own lockfiles based on what packages they have installed.

This pr renames occurrences of "Lockfile" to "Package". 
Also updates "lockfile_version" in schema to "manifest_version"

closes #73

### DIFF
--- a/spec/package.spec.json
+++ b/spec/package.spec.json
@@ -1,18 +1,18 @@
 {
-  "title": "Release Lock File Specification",
+  "title": "Package Specification",
   "type": "object",
   "required": [
-    "lockfile_version",
+    "manifest_version",
     "package_name",
     "version"
   ],
-  "version": "1",
+  "version": "2",
   "properties": {
-    "lockfile_version": {
+    "manifest_version": {
       "type": "string",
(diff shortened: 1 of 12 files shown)


spec/package.spec.json
```diff
@@ -0,0 +1,278 @@
...
+      "pattern": "^0x[0-9a-zA-Z]{64}$"
+    },
+    "IPFS-URI": {
+      "title": "An IPFS URI",
+      "type": "string",
+      "format": "uri",
+      "pattern": "^ipfs:/?/?.*$"
+    }
+  }
+}
```
gnidan: Should this also get rid of the old json spec?

njgheorghita: Yeah probably, I'll get rid of it now. I can't think of any reason we'd want to keep it around

gnidan:
Very in favor of this.

PR #91, opened 2018-05-11
Minor json schema tweaks

njgheorghita:
Suggesting a couple fixes to the json schema

### DIFF
--- a/spec/package.spec.json
+++ b/spec/package.spec.json
@@ -47,7 +47,7 @@
       "title": "The contract types included in this release",
       "type": "object",
       "patternProperties": {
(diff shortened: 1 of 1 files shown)


spec/package.spec.json
```diff
@@ -47,7 +47,7 @@
       "title": "The contract types included in this release",
       "type": "object",
       "patternProperties": {
-        "[a-zA-Z][-a-zA-Z0-9_]{0,255}(?:\\[[-a-zA-Z0-9]{1,256}\\])$": {
+        "[a-zA-Z][-a-zA-Z0-9_]{0,255}(?:\\[[-a-zA-Z0-9]{1,256}\\])?$": {
```
njgheorghita: I believe this `?` is necessary to make the second grouping optional - otherwise contract types without the second grouping will pass through without validation

PR #93, opened 2018-05-14
Migrate package spec description to Sphinx

gnidan:
- Convert pacakge.spec.md to RST automatically
- Rename Sphinx document
- Update link in README

ℹ️ [View this PR on the world wide web](http://ethpm.github.io/ethpm-spec/contrib/ethpm/gnidan/spec-rst/)


njgheorghita:
@gnidan nice! I like "package manifest specification" => "package manifest". Should we just get rid of "specification" altogether? like throughout `docs/source/package-spec.rst`

gnidan:
> Should we just get rid of "specification" altogether?

@njgheorghita I don't think it's redundant - the Package Manifest Specification specifies the format for package manifests. If anything, I'd suggest removing the word "manifest", since we refer to "package manifest" as just "package" in a bunch of places. I didn't do that here because this part of the documentation is low-level, so it seems prudent to be explicit.

njgheorghita:
@gnidan Ok I see your point, I'm not married to using "manifest", I think the only reason it stuck around was because "manifest_version" is used in `spec/package.spec.json`. So i'd be open to changing that to `spec_version` or something else? ( i think `package_version` might get a little confusing as individual packages will have their own versions) and deprecating "package manifest" for just "package" everywhere

gnidan:
Thanks for the approve! Let's discuss terminology in the channel or somewhere.

PR #70, opened 2017-11-09
start on defining link references

pipermerriam:
Fixes https://github.com/ethpm/ethpm-spec/issues/69

Work in progress.  TODO

PR #96, opened 2018-05-16
Better define link references

gnidan:
Update of #70 

Fixes https://github.com/ethpm/ethpm-spec/issues/69

### What was wrong

In the v1 version of the spec, link references were only defined as the locations within a block of bytecode in which an `__Thing____` placeholder was found.  This is problematic because:

1. It is imprecise, requiring regex based detection.
2. There are edge cases which cannot be handled once you start dealing with link values of varying lengths.

### How was it fixed

Now, there is a *Bytecode* object which defines three fields.

* `bytecode`: the actual bytecode itself
* `link_references`: the locations in the bytecode that require linking
* `link_dependencies`: the values that the bytecode should be linked with.

Link references are further defined in a *Link Reference* object which contains three fields:

* `offsets`: the start positions of the link reference
* `length`: for how long the link reference is
* `name`: for grouping of multiple link references which are expected to be linked with the same value. (optional)

All `bytecode` and `runtime_bytecode` fields have been replaced to use the *Bytecode* object definition instead.

ℹ️ [View preview](http://ethpm.github.io/ethpm-spec/contrib/ethpm/link-refs/)


examples/escrow/1.0.0.json
```diff
@@ -8,8 +8,12 @@
   },
   "contract_types": {
     "SafeSendLib": {
-      "bytecode": "0x606060405234610000575b60ad806100176000396000f36504062dabbdf050606060405260e060020a60003504639341231c81146024575b6000565b60306004356024356044565b604080519115158252519081900360200190f35b60003073ffffffffffffffffffffffffffffffffffffffff16318211156068576000565b60405173ffffffffffffffffffffffffffffffffffffffff84169083156108fc029084906000818181858888f19350505050151560a3576000565b5060015b9291505056",
-      "runtime_bytecode": "0x6504062dabbdf050606060405260e060020a60003504639341231c81146024575b6000565b60306004356024356044565b604080519115158252519081900360200190f35b60003073ffffffffffffffffffffffffffffffffffffffff16318211156068576000565b60405173ffffffffffffffffffffffffffffffffffffffff84169083156108fc029084906000818181858888f19350505050151560a3576000565b5060015b9291505056",
+      "bytecode": {
```
pipermerriam: Totally unrelated to this PR, but what do you think about renaming this to `deploy_bytecode`?  Removes the duplicate `bytecode` keys and adds semantic meaning as well as meaningful differentiation from `runtime_bytecode`

examples/escrow/1.0.0.json
```diff
@@ -117,10 +130,11 @@
         "address": "0x35b6b723786fd8bd955b70db794a1f1df56e852f",
         "transaction": "0x905fbbeb6069d8b3c8067d233f58b0196b43da7a20b839f3da41f69c87da2037",
         "block": "0x9b39dcab3d665a51755dedef56e7c858702f5817ce926a0cd8ff3081c5159b7f",
-        "link_dependencies": [
-          {"offset": 524, "value": "SafeSendLib"},
-          {"offset": 824, "value": "SafeSendLib"}
-        ]
+        "runtime_bytecode": {
+          "link_dependencies": [
+            {"offsets": [262, 412], "value": "SafeSendLib"}
```
pipermerriam: Some brief thoughts I'm having now about the `value` field.  Currently, we use three formats which are all distinctly distinguishable from each other

- `ContractName`
  - unique because it isn't one of the other two formats.
- `package-a:package-b:ContractName`
  - unique because of the embedded `:`
- `0xdeadbeef`
  - unique because it starts with a `0`

Should we just define this as it's own object with a `type` field

- `{'type': 'reference', 'value': 'SafeMathLib}`
- `{'type': 'literal', 'value': '0xdeadbeef'}`

My gut says yes because this pattern of regex based string differentiation has bit me before, and I *suspect* that it will again.  This specific case doesn't seem to have any edge cases which result in ambiguity, but it would be **very** easy for one to get introduced by accident without us realizing it.

Thoughts?

mkosowsk: I agree with going with an object for the `value` field as opposed to regex based string differentiation 👍🏻objects are more open for extension as well as easier to parse for people new to a codebase imo

examples/escrow/1.0.0.json
gnidan: Into this

gnidan: Actually how do you feel about `deployment_bytecode` to match word case (noun) with `runtime`?

examples/escrow/1.0.0.json
pipermerriam: Seem like we can just elevate it to the main object

```javascript
{"offsets": [262, 412], "type": "reference", "value": "SafeSendLib"}
```

gnidan:
@pipermerriam @mkosowsk please see recent commits and let me know if you have any other thoughts. Thanks!

mkosowsk:
@gnidan most recent commits pass smell test for me altho I don't have cycles to read for depth at the moment. LGTM 👍🏻

docs/source/package-spec.rst
```diff
@@ -355,14 +364,191 @@ packages that this project depends on.
...
+The ``value`` field defines the value which should be written when
+*linking* the corresponding bytecode.
+
+-  Required: Yes
+-  Type: String
+-  Format: determined based on ``type``:
+
+    Type ``literal``
+      For static value literals (e.g. address), value **must** be a
+      ``'0x'``-prefixed hexadecimal string with representing a whole number of
```
pipermerriam: I don't think the "representing a whole number of bytes" part should be here, but rather at the location where we define validity for hex encoded bytes.  Then we can just specifically say that the value must be a valid hex encoded byte string.

gnidan:
Merging. Thanks for the feedback!

PR #98, opened 2018-05-18
Begin glossary for docs

gnidan:
- Start to distinguish general concepts from *package spec*-specific information
- Use hyperlinks to glossary terms for first usage of terms in each section
- Use hyperlinks to spec format descriptions / conventions for usage / references for more detailed information.

This is an initial attempt to sort out docs organization practices; thoughts and opinions are appreciated!

ℹ️ [Preview](http://ethpm.github.io/ethpm-spec/contrib/ethpm/documentation)

### DIFF
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -1,0 +1,9 @@
+Glossary
+========
+
+.. include:: _include.rst
+
+.. glossary::
+
+   ABI
+      A JSON description of the external interface for a ABI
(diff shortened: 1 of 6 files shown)


docs/source/glossary.rst
```diff
@@ -0,0 +1,85 @@
+Glossary
+========
+
+.. include:: _include.rst
+
+.. glossary::
+
+   ABI
+      A JSON description of the external interface for a ABI
```
mkosowsk: Minor nitpick but think wording with something like this may be a bit more accessible to those new to scene:

PR #102, opened 2018-05-25
Update solidity and ipfs uris for transferable pkg

njgheorghita:
Update solidity version used in `transferable.sol`
Update ipfs links for `transferable` and `owned` pkg


njgheorghita:
@gnidan good nitpick - I was thinking about that, and brief scan of the style guide couldn't find - i'm down to capitalize filenames. should we also capitalize the package name then?

njgheorghita:
Actually - i'm good with capitalizing filename - and leaving package name uncapitalized. Not sure we need to push for a capitalized package name convention

gnidan:
Agreed that package names should be lowercase

njgheorghita:
mmmm. kebabs

davesque:
Did you notice my comment about the inconsistent indentation in `examples/transferable/1.0.0.json`?

PR #103, opened 2018-05-25
Update solidity for standard-token pkg

njgheorghita:
Updated solidity code & ipfs links for standard-token pkg


pipermerriam:
I *think* this is still a good example for a contract which is "ready to use", however, I mentioned to nick that these examples are going to be easiest for users to parse when the example contract we use is the "simplest semi-realistic demonstration of the concept".

So, for all of these, maybe take a minute to consider whether there is a *better* example, especially in any of these examples which have *large* contracts.  (but ERC20 is probably a good easy to understand example since it's well known among the development community.)

njgheorghita:
@davesque my bad, missed your comment about `transferable/1.0.0.json` - made the change here

PR #107, opened 2018-06-04
Update contracts for safe-math-lib

njgheorghita:
Contracts and manifest update for safe-math-lib package.

### DIFF
--- a/examples/safe-math-lib/1.0.0.json
+++ b/examples/safe-math-lib/1.0.0.json
@@ -3,49 +3,22 @@
   "version": "1.0.0",
   "package_name": "safe-math-lib",
   "sources": {
-    "./contracts/SafeMathLib.sol": "ipfs://QmVN1p6MmMLYcSq1VTmaSDLC3xWuAUwEFBFtinfzpmtzQG"
+    "./contracts/SafeMathLib.sol": "ipfs://QmNQMViTJ6dne4GNnTwHKnDwXxRaGAc5i2gNuU7T2G21uA"
   },
   "contract_types": {
     "SafeMathLib": {
       "deployment_bytecode": {
-        "bytecode": "0x606060405234610000575b60a9806100176000396000f36504062dabbdf050606060405260e060020a6000350463a293d1e88114602e578063e6cb901314604c575b6000565b603a600435602435606a565b60408051918252519081900360200190f35b603a6004356024356088565b60408051918252519081900360200190f35b6000828211602a57508082036081566081565b6000565b5b92915050565b6000828284011115602a57508181016081566081565b6000565b5b9291505056"
+        "bytecode": "0x604c602c600b82828239805160001a60731460008114601c57601e565bfe5b5030600052607381538281f30073000000000000000000000000000000000000000030146080604052600080fd00a165627a7a7230582083e86e46995b268179a35d3d89e28a4913a8104c518b63bc2b1aee04f9c0b6610029"
       },
       "runtime_bytecode": {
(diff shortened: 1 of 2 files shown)


docs/source/use-cases.rst
```diff
@@ -296,14 +296,15 @@ URI.
...
 
 This URI defines the chain on which the ``SafeMathLib`` library was
 deployed. The first hash you see,
 ``41941023680923e0fe4d74a34bdac8141f2540e3ae90623718e47d66d1ca4a2d`` is
 the genesis block hash for the Ropsten test network. The later hash
-``1e96de11320c83cca02e8b9caf3e489497e8e432befe5379f2f08599f8aecede`` is
-the block hash for block number 168,238 from the Ropsten chain.
+``5fd41fd8bbe935cdfc3bc89095b72dbd41c6308a4507e0f21059bf4a514995d4`` is
+the block hash for block number 3,372,354 from the Ropsten chain, and should
+represent any block on which the contract is available.
```
njgheorghita: @pipermerriam Am I remembering this correctly?

examples/safe-math-lib/1.0.0.json
```diff
@@ -3,80 +3,53 @@
...
-          ],
-          "name": "safeAdd",
-          "outputs": [
-            {"name": "","type": "uint256"}
-          ],
-          "payable": false,
-          "type": "function"
-        }
-      ],
+      "abi": [],
```
njgheorghita: ABI disappeared b/c the functions are now `internal` - therefore not part of the abi-interface -  which from what I can tell is library best practice.

docs/source/use-cases.rst
pipermerriam: Correct, though the language is a little awkward.  Maybe...

> ... from the Ropsten chain.  This block number was arbitrarily chosen as one *shortly* after block on which the contract was deployed.  Any block number from this chain for which the contract is available at that block would be valid here.

examples/safe-math-lib/1.0.0.json
pipermerriam: Hrm, that seems like something which is solidity specific.  It's probably inlining the actual function body into the contract that uses it which imho isn't an ideal extensible API since it would require that the compiler be able to compile the contract source.  This would disallow use of a Vyper library in a solidity contract or visa-versa.  I think these should remain externally callable.

examples/safe-math-lib/1.0.0.json
```diff
@@ -25,58 +25,60 @@
...
-              "b": "The amount that should be subtracted"
-            }
-          },
-          "safeSub(uint256,uint256)": {
             "details": "Adds a and b, throwing an error if the operation would cause an overflow.",
             "params": {
               "a": "The first number to add",
               "b": "The second number to add"
+            }
+		  },
```
davesque: Nitpick.  I think there are mixed tabs and spaces here that are throwing off the indentation.

davesque: Otherwise, I think this PR looks good.

njgheorghita: Good catch - do you know any handy vim tricks/plugins to deal with spaces/tabs? My vim always seems to cause a ruckus with them

davesque: I think you just need to make sure you have your `expandtab` option set:
```
:set expandtab
```
You could also make a `json.vim` file in your `ftplugin` directory that looks something like this:
```
set shiftwidth=2
set softtabstop=2
set tabstop=2
set expandtab
```

PR #108, opened 2018-06-04
Update escrow package resources

njgheorghita:
Update escrow package to use a newer version of Solidity


examples/escrow/1.0.0.json
```diff
@@ -52,10 +53,10 @@
...
         "link_references": [
           {
-            "offsets": [262, 412],
+            "offsets": [301, 495],
```
gnidan: What are those extra `0`s blocks doing?

njgheorghita: Hard for me to tell - my opcode game isn't solid. But that's the `runtime_bytecode` output when remix compiles the contract - and they're also there if you check the runtime code on Etherscan. I don't think they have anything to do with our link references... 🤷‍♂️

PR #114, opened 2018-06-12
Minify pkgs & create pretty versions

### LINKED ISSUE #105: Ensure examples' package manifests are properly stripped of whitespace and sorted
Since #77, this spec requires packages to be minified, but this repo still only provides the pretty-printed version in the `examples/` directories.

I propose that we keep two copies of each example's JSON, one minified and sorted, the other kept with indentation and semantically-ordered object keys. This change can happen in one of two directions:
a. `1.0.0.json` for strict, `1.0.0-pretty.json` for pretty
b. `1.0.0.json` for pretty, `1.0.0-strict.json` for strict

Opening this issue to gauge others' thoughts. On the one hand, there's the risk of external links going bad, replaced with unreadable JSON content, but on the other hand, the strict form _is_ canonical. I slightly prefer (a); other thoughts?

Thanks!

njgheorghita:
- Rename `1.0.0.json` -> `1.0.0-pretty.json`
- Sort and minify pkgs -> save as `1.0.0.json`

fixes #105


gnidan:
Ah shoot @njgheorghita I had to do this separately. I just pushed a `pre-minify` branch with only the renames. What command did you run to minify, so I can do that? Alternatively, want to redo it yourself so you don't lose commit credit?

njgheorghita:
@gnidan No worries - I have a little script to minify/prettify so I just went ahead and did it again. Though - I did `git mv` all the files before creating the minified versions and it seems as though git is still not recognizing the rename? Possibly because it recognizes `1.0.0.json` as being changed to strict - not renamed to `-pretty.json`

gnidan:
Ahhh! Git is living up to its name. Oh well. Goodbye `git blame` :)

PR #121, opened 2018-08-15
Use `description` instead of `title`

gnidan:
Since most `title` properties serve as descriptions.

For top-level definitions, provide `title` with values that can serve as a variable name


gnidan:
Thanks Piper!

PR #122, opened 2018-08-21
Fix a couple schema titles + required fields

gnidan:
- Force address to be required for instances
- Distinguish `ContractTypeName` from `ContractType`
- Rename `ContractBytecode` to `BytecodeObject`

### DIFF
--- a/spec/package.spec.json
+++ b/spec/package.spec.json
@@ -166,7 +166,7 @@
       ],
       "properties": {
         "contract_type": {
-          "title": "Contract Type",
+          "title": "Contract Type Name",
           "description": "The contract type of this contract instance",
           "type": "string",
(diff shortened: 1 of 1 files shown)


gnidan:
(Ah, was careless. Fixing)

spec/package.spec.json
```diff
@@ -161,11 +161,12 @@
       "description": "Data for a deployed instance of a contract",
       "type": "object",
       "required": [
-        "contract_type"
+        "contract_type",
+        "address"
```
njgheorghita: Nice catch!

gnidan: It's all TypeScript saving my bacon :bacon:

gnidan:
Thanks!

Issue #123, opened 2018-08-21
Fix BytecodeObject validity requirements


gnidan:
Noticed a couple issues with BytecodeObject's definition, or inconsistencies with the natural language spec:

- BytecodeObject is defined in the JSON-Schema to require either `bytecode` or `link_dependencies`, but not both
- The natural language spec indicates that `bytecode` is always required, and `link_dependencies` is never required.

Presumably, `link_dependencies` is indicated as required in the JSON-Schema because it is necessary for ContractInstanceObject definitions. This means that the spec expects to see `link_dependencies: []` when no link references are present. Similarly for contract instances, `bytecode` can be omitted, inheriting from the contract type.

I propose that we change these definitions a bit, splitting `BytecodeObject` into the non-polymorphic:
- `UnlinkedBytecodeObject`
- `LinkedBytecodeObject`

`UnlinkedBytecodeObject` would require both `bytecode` and `link_references` (empty array for no links). `LinkedBytecodeObject` would formally require _only_ `link_dependencies`, but semantically require `bytecode` and `link_references` if and only if they differ from inherited values.

We would then update `ContractType` and `ContractInstance` to use unlinked / linked forms, respectively.

Thoughts? I opened a PR with formal schema changes, linked below.

gnidan:
Oh, another possibly open question:

Does a contract type that is an alias for another type need to include `bytecode`?

pipermerriam:
> Does a contract type that is an alias for another type

It's not clear to me what you mean by this.

gnidan:
@pipermerriam Maybe I don't understand the use case for [Contract Aliases](http://ethpm.github.io/ethpm-spec/glossary.html#term-contract-alias). Presumably the purpose is to define multiple contract types for various "singleton" roles? So one contract type would be canonical, the others being aliases. In such a situation, bytecode can potentially be omitted for alias types.

njgheorghita:
Yeah ok, I think I like this proposal, but it's not yet sold that creating more explicit bytecode classes is a useful distinction, or an unnecessary complication. Do the benefits outweigh just correcting/updating the natural language spec? 

> This means that the spec expects to see link_dependencies: [] when no link references are present.

But wouldn't there be an accompanying `bytecode` so the spec doesn't expect to see (aka require) `link_dependencies` at all?

> but semantically require bytecode and link_references if and only if they differ from inherited values.

What do you mean by semantically? It's just up to the manifest creator to include these fields if they differ?

[...]

PR #124, opened 2018-08-22
[proposed] Update BytecodeObject validation requirements

gnidan:
As proposed in #123.

This contains only the formal spec changes. Will need to rewrite the natural language spec if this is change has support.


gnidan:
Thanks @njgheorghita! If you want to take a stab at the docs, be my guest. I won't have time before tomorrow to get to it, unfortunately. Or just feel free to tell me it's my responsibility :)

PR #125, opened 2018-08-22
Add first-class Deployment definition

gnidan:
To improve schema accessibility for tools that do schema processing, etc.

This arose from using `json-schema-to-typescript`; reducing nesting and giving things names improves types generation.


gnidan:
Thanks!

PR #126, opened 2018-08-24
Update bytecode object in docs

njgheorghita:
Natural language spec update for proposed #123 

Hope I'm understanding this correctly, very open to all suggested changes/feedback to improve understanding/clarity.

### DIFF
--- a/docs/source/package-spec.rst
+++ b/docs/source/package-spec.rst
@@ -405,10 +405,10 @@ within the dependency tree the value is constructed as follows.
 
 ----
 
-The *Bytecode* Object
-~~~~~~~~~~~~~~~~~~~~~
+The *UnlinkedBytecode* Object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-A bytecode object has the following key/value pairs.
+An unlinked bytecode object, representing a |ContractType|, has the following key/value pairs.
 
 Bytecode: ``bytecode``
 ^^^^^^^^^^^^^^^^^^^^^^
@@ -420,14 +420,13 @@ hexadecimal representation of the bytecode.
   :Type: String
   :Format: ``0x`` prefixed hexadecimal.
 
-
 Link References: ``link_references``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 The ``link_references`` field defines the locations in the corresponding
-bytecode which require |linking|.
+bytecode which require |linking|. If bytecode contains no link references, this field should be an empty array.
 
-  :Required: No
+  :Required: Yes
   :Type: Array
   :Format: All values **must** be valid :ref:`Link Reference objects <link-reference-object>`.
     See also below.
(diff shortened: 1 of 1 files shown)


pipermerriam:
this is great, because now I can see what I think is a problem...  As it stands:

- contract instances: linked bytecode
- contract type: unlinked bytecode

I think this ?unintentionally? removes the option to ship contract types with their bytecode pre-linked.  It may not be explicitely stated, and it probably should be explicitely stated but I think this is an important use case to support.  Without this, it becomes impossible to ship ready-to-use contracts which make use of libraries.

Am I missing something?

docs/source/package-spec.rst
```diff
@@ -420,14 +420,13 @@ hexadecimal representation of the bytecode.
   :Type: String
   :Format: ``0x`` prefixed hexadecimal.
 
-
 Link References: ``link_references``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 The ``link_references`` field defines the locations in the corresponding
-bytecode which require |linking|.
+bytecode which require |linking|. If bytecode contains no link references, this field should be an empty array.
```
pipermerriam: s/should/must/

gnidan:
> I think this ?unintentionally? removes the option to ship contract types with their bytecode pre-linked. It may not be explicitely stated, and it probably should be explicitely stated but I think this is an important use case to support. Without this, it becomes impossible to ship ready-to-use contracts which make use of libraries.

I'd say that if a package doesn't want to expose already-linked types, it can just say `link_references: []`. This doesn't change the underlying unlinkedness of the object, it just means that a particular piece of bytecode doesn't have any public link references.

njgheorghita:
> I'd say that if a package doesn't want to expose already-linked types, it can just say link_references: []. 

I might be missing something, but does this cover if a contract type does want to expose an already-linked library (maybe for validation purposes), yet still have unlinked bytecode/link references?

gnidan:
> I might be missing something, but does this cover if a contract type does want to expose an already-linked library (maybe for validation purposes), yet still have unlinked bytecode/link references?

Hm, for validation purposes? This is probably only covered via the use of the compiler settings property, to indicate that specific link values were provided to the compiler ahead of time.

njgheorghita:
@pipermerriam Added `link_dependencies` option to `UnlinkedBytecodeObject`. @gnidan I'd appreciate a final look over to make sure everything looks good

gnidan:
Another thing: what if a contract only wants to pre-link _some_ of its link refs? So then `link_dependencies` would have a subset of values and not have the same array length as `link_references`.

njgheorghita:
Feels weird to convert the linked bytecode value to `0`s. 

Can we just create a new "type" of link value?

njgheorghita:
Is it a requirement that `link_dependencies` and `link_references` are the same length? As long as all `link_dependencies` reference a `link_reference`, it should be fine to have `link_references` that don't reference a `link_dependencies`? 🤷‍♂️

gnidan:
And we can change the zeroes requirement from **must** to **should**, that way we can try to get the tools to zero out pre-fill values, but it's not entirely necessary, really.

But now I'm unsure of the distinction that types are unlinked and instances are linked. Is there even a useful distinction there anymore? All we know is that instances **must** have values for all link references.

njgheorghita:
It also might be somewhat confusing to have an `UnlinkedBytecodeObject` that can actually be (partially) linked in certain cases. A better name could help though

Hackdom:
So after reading through this and the spec and my code over and over, here's what seems to make sense in my brain. We go back to a single bytecode object (this is where offsets live anyway) and require all links be listed in a mapping using the offset as key. These links can then be one of three types: placeholder, literal, or reference. Then from these types we can determine the rest of the information as follows:
```
bytecode_object: {
  bytecode: [string always required],
  links: {
    [int defining offset]: {
      type: ['placeholder','literal','reference'],
      // if 'placeholder' then value is the placeholder name, if 'literal' then value is the literal string, if 'reference' then value is the reference name
      value: [required string], 
      length: [required int if type is 'placeholder'],
    } 
  }
}
```

PR #133, opened 2018-10-03
Use e.g. (for example) instead of i.e. (in essence)

davidmurdoch:
"e.g." should be used when providing an example, whereas "i.e." should be use when providing clarification.

### DIFF
--- a/docs/source/_use-cases.rst
+++ b/docs/source/_use-cases.rst
@@ -5,42 +5,42 @@ The following use cases were considered during the creation of this
 specification.
 
 Stand Alone Package with an Inheritable Contract
-   i.e. ``owned``
+   e.g. ``owned``
 
    See `full description <#stand-alone-package-with-an-inheritable-contract>`__.
 
 Dependent Package with an Inheritable Contract
-   i.e. ``transferable``
+   e.g. ``transferable``
 
    See `full description <#dependent-package-with-an-inheritable-contract>`_.
 
 Stand Alone Package with a Reusable Contract
-   i.e. ``standard-token``
+   e.g. ``standard-token``
 
    See `full description <#stand-alone-package-with-a-reusable-contract>`_.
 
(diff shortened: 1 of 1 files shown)


gnidan:
Huh, I think we meant to say "id est" because those are the names of the use cases?

pipermerriam:
Things I've never taken the time to fully internalize.

- it's its
- e.g. i.e.
- affect effect
- general spelling
- not leaving snarky remarks on pull requests.

Here's a squirrel eating a pumpkin

squirrel in pumpkin photograph 2014 thanksgiving photo - 2014 thanksgiving squirrel in pumpkin autum-f41515

davidmurdoch:
@gnidan, when I first read it I thought of the use case names (owned, transferable, etc.) as examples of the use case description titles.

So the "owned" contract was an example of a "Stand Alone Package with an Inheritable Contract".

I can see now what your intent was.

I have no issue if you want to close this PR. Perhaps the first section could just be a simple Table of Contents as a list, with each full description section linking to the folder of its example implementation. Let me know if this is something you'd prefer and I can open a PR real quick (or not) .

Words are hard.


@pipermerriam Are we sure that the pumpkin isn't eating the squirrel though? 🎃

njgheorghita:
I'm getting horrible flashbacks to middle school english class - but according to [this](https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/)
```
E.g. is used to give one or more possible examples. It’s a signal that you’re seeing one or a few of multiple possibilities.

I.e, on the other hand, clarifies; you are providing more precise information. Where e.g. opens up more options, i.e. narrows them down.
``
It seems to me that `e.g.` is more appropriate. That being said, I really don't think it makes a big difference either way - we are programmers after all not literature experts - so I'm gonna move ahead and merge this - pls ping me if you think this was the wrong move :heart:

PR #135, opened 2018-12-02
Update JS package info / hooks

gnidan:
- Auto-generate TypeScript interfaces from spec
- Remove old docs generator script
- Fix package.json repo reference
- Update IPFS URIs


.gitignore
```diff
@@ -1,9 +1,11 @@
 node_modules
-yarn.lock
```
njgheorghita: Is there a specific reason we want to include `yarn.lock`? I'm just not too familiar with yarn...

gnidan: Yep, the [recommendation](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/) is to commit this to source control

PR #136, opened 2019-05-13
Update "URI" format in schema

njgheorghita:
Format `"URI"` is undefined, and should be `"uri"`.

```
unknown format "URI" is used in schema at path "#/definitions/PackageMeta/properties/links/additionalProperties"
```

Also updated some dependencies that had security vulnerabilities

### DIFF
--- a/docs/source/_use-cases.rst
+++ b/docs/source/_use-cases.rst
@@ -12,37 +12,37 @@ Stand Alone Package with an Inheritable Contract
 Dependent Package with an Inheritable Contract
    i.e. ``transferable``
 
-   See `full description <#dependent-package-with-an-inheritable-contract>`_.
+   See `full description <#dependent-package-with-an-inheritable-contract>`__.
 
 Stand Alone Package with a Reusable Contract
    i.e. ``standard-token``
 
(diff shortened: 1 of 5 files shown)


pipermerriam:
Any thoughts on a small test suite that validates the schema against known good/bad examples?

njgheorghita:
@pipermerriam We validate the schema against valid manifests [here](https://github.com/ethpm/ethpm-spec/blob/master/bin/validate.py) - and there is more testing done in py-ethpm [here](https://github.com/ethpm/py-ethpm/blob/3a9a60670ea8143938a32a0fc982d612747ef214/tests/ethpm/utils/test_manifest_validation_utils.py) and [here](https://github.com/ethpm/py-ethpm/blob/8944dfa4f844020bec122c6f0e21a705a90271e3/tests/ethpm/test_manifest_assets_are_valid.py). 

The bug fixed in this pr was only raised now because I just added a new tool in [ethpm.js](https://github.com/ethpm/ethpm.js/pull/25) for json schema validation - so I guess the python tooling just didn't pick up on it? 

It's possible more invalid manifests should be tested against, but I'm not sure how far to take it or where to store them as there are many many possible invalid manifests, and clogging this library up with them might be distracting / confusing :shrug:

PR #141, opened 2020-03-27
V3

njgheorghita:
The changes to the V2 spec discussed in the stakeholders sync yesterday. These changes are not finalized and are open for suggestions / amendments.

### DIFF
--- a/docs/source/package-spec.rst
+++ b/docs/source/package-spec.rst
@@ -89,63 +89,67 @@ name collisions with future versions of the specification.
 
 ----
 
-EthPM Manifest Version: ``manifest_version``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+EthPM Manifest Version: ``manifest``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The ``manifest_version`` field defines the specification version that
-this document conforms to. Packages **must** include this field.
+The ``manifest`` field defines the specification version that
+this document conforms to.
+
+  - Packages **must** include this field.
 
   :Required: Yes
-  :Key: ``manifest_version``
+  :Key: ``manifest``
   :Type: String
-  :Allowed Values: ``2``
+  :Allowed Values: ``ethpm/3``
 
-.. _package names:
+.. _package-names:
 
 ----
 
-Package Name: ``package_name``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Package Name: ``name``
+~~~~~~~~~~~~~~~~~~~~~~
 
-The ``package_name`` field defines a human readable name for this
-package. Packages **must** include this field. Package names **must**
-begin with a lowercase letter and be comprised of only lowercase
-letters, numeric characters, and the dash character ``-``. Package
-names **must** not exceed 255 characters in length.
+The ``name`` field defines a human readable name for this package.
 
-  :Required: Yes
-  :Key: ``package_name``
+  - Packages **must** include this field to be released on an EthPM registry.
+  - Package names **must** begin with a lowercase letter and be comprised of only lowercase
+    letters, numeric characters, and the dash character ``-``.
+  - Package names **must** not exceed 255 characters in length.
+
+  :Required: If ``version`` is included.
+  :Key: ``name``
   :Type: String
   :Format: **must** match the regular expression
-    ``^[a-z][a-z0-9_-]{0,255}$``
+    ``^[a-z][-a-z0-9]{0,255}$``
 
 ----
 
-Package Meta: ``meta``
-~~~~~~~~~~~~~~~~~~~~~~
+Package Version: ``version``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``version`` field declares the version number of this release.
+  
+  - Packages **must** include this field to be released on an EthPM registry.
+  - This value **should** conform to the `semver <http://semver.org/>`__ version numbering specification.
+
+  :Required: If ``name`` is included.
+  :Key: ``version``
+  :Type: String
+
+----
+
+Package Metadata: ``meta``
+~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The ``meta`` field defines a location for metadata about the package
 which is not integral in nature for package installation, but may be
-important or convenient to have on-hand for other reasons. This field
-**should** be included in all Packages.
+important or convenient to have on-hand for other reasons.
+
+  -  This field **should** be included in all Packages.
 
   :Required: No
   :Key: ``meta``
   :Type: `Package Meta Object`_
-
-----
-
-Version: ``version``
-~~~~~~~~~~~~~~~~~~~~
-
-The ``version`` field declares the version number of this release. This
-value **must** be included in all Packages. This value **should**
-conform to the `semver <http://semver.org/>`__ version numbering
-specification.
-
-  :Required: Yes
-  :Key: ``version``
-  :Type: String
 
 ----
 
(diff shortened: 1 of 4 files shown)


njgheorghita:
ping @iamdefinitelyahuman if you wouldn't mind giving this a quick :eyes: to verify i essentially captured the conclusions of the call - then i'll push this to our docs for a wider review from everybody else

docs/source/package-spec.rst
```diff
@@ -89,63 +89,67 @@ name collisions with future versions of the specification.
...
-The ``package_name`` field defines a human readable name for this
-package. Packages **must** include this field. Package names **must**
-begin with a lowercase letter and be comprised of only lowercase
-letters, numeric characters, and the dash character ``-``. Package
-names **must** not exceed 255 characters in length.
+The ``name`` field defines a human readable name for this package.
 
-  :Required: Yes
-  :Key: ``package_name``
+  - Packages **must** include this field to be released on an EthPM registry.
```
pipermerriam: I'm not sure about this line.  I think i see the motivation here, but I'm conflicted about where the right place for this should be and I don't think it is here.  

| | named | unnamed |
| --- | --- | --- |
| **from registry** | must match | default to name from registry |
| **not from registry** | always valid | user must provide name |

I'd propose going with the logic above which suggests that the name is strictly optional but providing a warning that a package without a name is subject to ambiguous naming.

docs/source/package-spec.rst
```diff
@@ -154,52 +158,43 @@ Sources: ``sources``
...
-  :Format:
-      Keys **must** be valid |ContractAliases|.
+The ``compilers`` field holds the information about the compilers and their
+settings that have been used to generate the various ``contractTypes`` included
+in this release.
 
-      Values **must** conform to the `Contract Type Object`_ definition.
+  :Required: No
+  :Key: ``compilers``
+  :Type: Array (`the Compiler Information object`_)
```
pipermerriam: I'm confused by this being an array.  Still reading so maybe it will be self explanitory.

docs/source/package-spec.rst
```diff
@@ -518,13 +505,69 @@ for the following common resources.
...
 ----
 
+.. _Sources Object:
+
+The *Sources* Object
+~~~~~~~~~~~~~~~~~~~~
+
+A *Sources* object is defined to have the following fields.
+
+  :Key: A global identifier for the source file. (string)
```
pipermerriam: I think this suggests that all keys should be unique but doesn't state it explicitly.  Maybe it is worth having a global rule that all objects **must** have a unique set of keys unless otherwise specified.

docs/source/package-spec.rst
```diff
@@ -518,13 +505,69 @@ for the following common resources.
...
+Source Object
+^^^^^^^^^^^^^
+
+Checksum: ``checksum``
+^^^^^^^^^^^^^^^^^^^^^^
+
+Hash of the source file - prefixed by the hash algorithm and a ``/``.
+  - (eg. ``keccak256/abc123...``).
+
+  TODO: Can we enforce content-addressed urls and remove this field entirely?
```
pipermerriam: just noting the `TODO` still in here.

docs/source/package-spec.rst
```diff
@@ -518,13 +505,69 @@ for the following common resources.
...
+.. _SourceObject:
+
+Source Object
+^^^^^^^^^^^^^
+
+Checksum: ``checksum``
+^^^^^^^^^^^^^^^^^^^^^^
+
+Hash of the source file - prefixed by the hash algorithm and a ``/``.
+  - (eg. ``keccak256/abc123...``).
```
pipermerriam: I'd be inclined to move this to be an object with two fields rather than depending on string interpolation which is inherently imprecise and often leads to unexpected ambiguity.

docs/source/package-spec.rst
```diff
@@ -518,13 +505,69 @@ for the following common resources.
...
+
+  TODO: Can we enforce content-addressed urls and remove this field entirely?
+  :Required: If there are no urls that contain a hash of its contents.
```
pipermerriam: Maybe reword as:

?  Required **if** the the URI does not contain a content hash.

docs/source/package-spec.rst
```diff
@@ -518,13 +505,69 @@ for the following common resources.
...
+  :Key: ``checksum``
+  :Value: string
+
+URLs: ``urls``
+^^^^^^^^^^^^^^
+Array of urls that resolve to the same source file.
+  - Urls **should** be stored on a content-addressable filesystem.
+  - If the resulting document is a directory the key **should** be interpreted as a directory path.
+  - If the resulting document is a file the key **should** be interpreted as a file path.
```
pipermerriam: Do we want anything in here about the URLs requiring a proper `scheme`  i.e. we probably don't want: `QmPXME1oRtoT627YKaDPDQ3PwA8tdP9rWuAAweLzqSwAWT` but rather want `ipfs://QmPXME1oRtoT627YKaDPDQ3PwA8tdP9rWuAAweLzqSwAWT`

docs/source/package-spec.rst
```diff
@@ -518,13 +505,69 @@ for the following common resources.
...
+
+  :Required: If ``urls`` is not present.
+  :Key: ``content``
+  :Value: string
+
+Install Path: ``installPath``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Filesystem path of source file.
+  - **Must** be a relative filesystem path that begins with a ``./``.
+  - **Must** resolve to a path that is within the current virtual working directory.
```
pipermerriam: I think this is missing a validity rule that:

> **Must** be unique across all included sources.

docs/source/package-spec.rst
```diff
@@ -746,6 +778,18 @@ to the `Compiler Input and Output Description <http://solidity.readthedocs.io/en
   :Key: ``settings``
   :Type: Object
 
+Contract Types: ``contractTypes``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A list of the |ContractAlias| in this package that used this compiler to generate its outputs. 
+
+  - All ``contractTypes`` that locally declare ``runtimeBytecode`` **should** be attributed for by a compiler object.
+  - A single ``contractTypes`` **must** not be attributed to more than one compiler.
```
pipermerriam: I think I see here is why the compiler bit is an array yes?

docs/source/package-spec.rst
njgheorghita: Definitely - we have a rule like that already defined in the document formatting section: `Duplicate keys in the same object are invalid.`.

docs/source/package-spec.rst
njgheorghita: Gotcha, agreed this seems like a neater way of handling this.

docs/source/package-spec.rst
njgheorghita: @pipermerriam Yeah, the general conclusion was that the `compiler` info is better suited as an input, when the rest of `contractTypes` info is more of an output (`abi`, `bytecodes`, etc), so it would be cleaner to separate the two. Especially when it comes to using packages as compiler json inputs - considering also packages with `.vy` & `.sol` sources - a compiler doesn't have to parse all the sources to figure out the relevant files.

There are also some DRY benefits - eg. when many `contractTypes` share the same compiler.

One question I have is - along the lines of DRY. I'm trying to think if it's reasonable to drop `compiler` / `runtime_bytecode` from a `deployment` - and if a user wants to do any validation require that a corresponding `contractType` is available. But, this could also just result in an excessive amount of duplication in `contractTypes` to account for many deployments w/ different `linkReferences`.

pipermerriam: I'm sorry but I don't follow your question.  Can you pose it with a simple example maybe of the before/after?

njgheorghita: Sure! Essentially I'm trying to figure out if we can remove the `"compiler"` object from  a `"deployments"` object.

v2: Compilers can be defined within a deployment. The spec requires that a `"deployment"` defines a `"contract_type"` and that the specified `"contract_type"` (eg `"Escrow"`) is defined within the package's top-level `"contract_types"` - which also can contain a `"compiler"` object
```json
{
  "contract_types": {
    "Escrow": {
      "compiler": {...}
    }
  },
  "deployments": {
    "blockchain://abc/block/123": {
     "Escrow": {
        "contract_type": "Escrow",
        "address": "0x8c1...",
        "transaction": "0xc60...",
        "block": "0x4d1a...",
        "runtime_bytecode": {
          "link_dependencies": [
            {"offsets": [301, 495], "type": "reference", "value": "SafeSendLib"}
          ]
        },
        "compiler": {
          "name": "solc",
          "version": "0.4.24+commit.e67f0147.Emscripten.clang",
          "settings": {
            "optimize": true
          }
        }
      }
  }
}
```

v3: Moving compiler information to a top-level `"compilers"`, this should be sufficient to match a compiler to all `"contractTypes"` - and their respective `"deployments"`. Allowing us to remove the `"compiler"` object from within a `"deployments"` object.
```json
{
  "compilers": [{
    "name": "solc",
    "version": "0.4.24....",
    "contractTypes": ["Escrow"],
  }]
}
```

Let me know if that helps - or more clarification is needed.

pipermerriam: :+1:

PR #144, opened 2020-04-28
Display both v2 and v3 package specifications

njgheorghita:
The spec documentation was only showing the v3 proposed changes. For reference for upcoming discussions, it's important that v2 is easily accessible. Unfortunately github pages doesn't let you deploy multiple versions from different branches, so I just broke them out into separate categories here.


njgheorghita:
merging since this is a simple rearrangement of docs

PR #145, opened 2020-05-06
Update source object and contract alias definition

njgheorghita:
Proposed changes from the Solidity Summit ethpm session.

### DIFF
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -64,11 +64,8 @@ Glossary
 
       The contract alias **must** use *one of* the following naming schemes:
 
-      -  ``<source-id>:<contract-name>``
-      -  ``<source-id>:<contract-name>[<identifier>]``
-
-      The ``<source-id>`` portion **must** match a key from the ``sources``
-      object for the respective source file of this contract type.
+      -  ``<contract-name>``
+      -  ``<contract-name>[<identifier>]``
 
       The ``<contract-name>`` portion **must** be the same as the
       |ContractName| for this contract type.
--- a/docs/source/v3-package-spec.rst
+++ b/docs/source/v3-package-spec.rst
@@ -465,10 +465,10 @@ authors of this package. Packages **may** include this field.
 License: ``license``
 ^^^^^^^^^^^^^^^^^^^^
 
-The ``license`` field declares the license under which this package is
-released. This value **should** conform to the
+The ``license`` field declares the license associated with this
+source file. This value **should** conform to the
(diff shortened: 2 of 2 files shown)


pipermerriam:
It would be nice to have an explanation of these changes here, even if only for historical purposes.

docs/source/glossary.rst
```diff
@@ -64,11 +64,8 @@ Glossary
 
       The contract alias **must** use *one of* the following naming schemes:
 
-      -  ``<source-id>:<contract-name>``
-      -  ``<source-id>:<contract-name>[<identifier>]``
-
-      The ``<source-id>`` portion **must** match a key from the ``sources``
-      object for the respective source file of this contract type.
+      -  ``<contract-name>``
+      -  ``<contract-name>[<identifier>]``
```
njgheorghita, line 68: The idea to prefix a contract alias with `<source-id>` was reversed. Since source id's have no restrictions on characters, using a `:` to parse contract aliases (particularly when identifying a contract type of a dependency) is very difficult to handle.

docs/source/v3-package-spec.rst
```diff
@@ -566,6 +566,16 @@ Filesystem path of source file.
...
+^^^^^^^^^^^^^^
+
+The ``type`` field declares the type of the source file. The field **should** be one
+of the following values: ``jsonabi``, ``solidity``, ``vyper``. 
+
+  :Required: No
+  :Key: ``type``
+  :Type: String
+
 ----
```
njgheorghita, line 579: A `type` field was added to `SourceObject` to make it simpler for compilers to identify what files they should be concerned about.

docs/source/v3-package-spec.rst
```diff
@@ -646,6 +656,14 @@ Natspec: ``natspec``
     `UserDoc <https://github.com/ethereum/wiki/wiki/Ethereum-Natural-Specification-Format#user-documentation>`_
     and `DevDoc <https://github.com/ethereum/wiki/wiki/Ethereum-Natural-Specification-Format#developer-documentation>`_  formats.
 
+Source: ``source``
+^^^^^^^^^^^^^^^^^^
+The global source identifier for the source file from which this contract type was generated.
+
+  :Required: No
+  :Type: String
+  :Value: **must** match a unique source ID included in the `Sources Object`_ for this package.
```
njgheorghita, line 665: This `source` field was added to compensate for the change in a contract alias mentioned above. Contract types simply point to their source via this field.

PR #146, opened 2020-05-08
Final updates before public last call for v3 freeze

njgheorghita:
Final changes from sync implemented into docs and schema.

Includes...
- Remove requirement for brackets to surround contract alias identifiers. It was felt that this was an unnecessarily strict requirement.
- Update contract type name regex to include `_` and `$` which are valid solidity contract names. 
-  Update docs to reflect `natspec` -> `userdoc` & `devdoc`
-  Add `source` field to a contract type object to point towards the respective source file.
- Some refactoring in schema to reduce duplicated regex
- update v3 examples

TODO (in future prs):
- comprehensive tests for machine readable schema / regex


spec/v3.spec.json
```diff
@@ -67,8 +66,11 @@
     "buildDependencies": {
       "title": "Build Dependencies",
       "type": "object",
+      "propertyNames": {
+        "$ref": "#/definitions/PackageName"
+	  },
       "patternProperties": {
-        "^[a-z][-a-z0-9]{0,255}$": {
+        "": {
```
pipermerriam, line 73: This must be JSON-schema syntax that I'm not familiar with.  Am I correct that the previous `"propertyNames"` definition gets applied to the keys for `"patternProperties"`?

spec/v3.spec.json
```diff
@@ -96,13 +98,18 @@
...
 		  "type": "string"
 		},
 		"installPath": {
 		  "title": "Target file path to install source file",
 		  "type": "string",
 		  "pattern": "^\\.\\/.*$"
+		},
+		"type": {
+		  "title": "Type",
+		  "description": "File type of this source",
```
pipermerriam, line 111: Is this a new field in the spec?  What types of values would go here?  What is this used for?

spec/v3.spec.json
```diff
@@ -159,6 +172,11 @@
           "type": "string",
           "pattern": "[a-zA-Z][a-zA-Z0-9_]{0,255}"
         },
+        "source": {
```
pipermerriam, line 175: It might be nice if this key was something like `sourceId` or `sourceIdentifier` or something like that to make it abundantly clear that this isn't the actual source code but rather a pointer to it.

spec/v3.spec.json
njgheorghita: Yup, it was added in a [newer version of jsonschema](https://json-schema.org/understanding-json-schema/reference/object.html#property-names).

docs/source/glossary.rst
```diff
@@ -65,13 +67,13 @@ Glossary
       The contract alias **must** use *one of* the following naming schemes:
 
       -  ``<contract-name>``
-      -  ``<contract-name>[<identifier>]``
+      -  ``<contract-name><identifier>``
```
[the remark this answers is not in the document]
njgheorghita: A `ContractTypeObject` has a `contractName` field which is required if the contract alias differs from the contract name. So the `<contract-name>` is easily available information.

spec/v3.spec.json
njgheorghita: [Yup](http://ethpm.github.io/ethpm-spec/v3-package-spec.html#id2).

> The type field declares the type of the source file. The field should be one of the following values: `jsonabi`, `solidity`, `vyper`.

This was a request from the compiler teams since it makes it simple for a compiler to know which files they should care about / ignore.

PR #147, opened 2020-05-11
Write tests for v3 jsonschema

njgheorghita:
- Write tests to validate jsonschema is working as expected.
- Add test run to travis build

@pipermerriam 
It appears as though in V2 - a deployment can only refer to a contract type from a direct dependency.
```json
{
  "buildDependencies": {
    "mydependency1": "ipfs://Qm..."
  },
  "deployments": {
    "blockchain://...": {
      "MyContract": {
        "contractType": "mydependency1:MyContract", // valid
        "contractType": "mydependency1:mydependency2:MyContract" // invalid
      }
    }
  }
}
```

Is there a reason why it's restricted to only reference contract types in a direct dependency - rather than multiple levels? We allow this kind of nested referencing for a [link value] - but is there a reason why we shouldn't allow this for a deployment's contract type?


pipermerriam:
> Is there a reason why it's restricted to only reference contract types in a direct dependency - rather than multiple levels?

No reason I can remember or think of.  Seems like a bug/error in the spec.

tests/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+from jsonschema import validate
+from jsonschema.validators import (
+    Draft7Validator,
+    validator_for,
+)
+
+
+@pytest.fixture
+def schema():
+    p = Path('.')
```
pipermerriam, line 13: Why not just instantiate the whole path here?

Also, you probably shouldn't make assumptions about what the current directory is.  The more robust way to do this is anchoring to `__file__`


```python
ROOT = Path(__file__).parent
path = ROOT / <whatever>
```

This still depends on the location of the actual python module but it is more reliable.

tests/standardized/meta/invalid/invalidAuthors0.json
```diff
@@ -0,0 +1,8 @@
+{
+	"testInfo":{
+		"testCase": "invalid",
+		"errorPointer":"['meta']['authors']",
+		"reason": "'mr. author' is not of type 'array'"
+	},
+	"package":{"manifest":"ethpm/3","meta":{"authors":"mr. author"}}
```
njgheorghita, line 7: @pipermerriam Any thoughts regarding on this structure for the wrapper? From what I read, I couldn't find any kind of "universal" error codes, but it seems like the erroneous json pointer should be capable of that role. It's also unlikely that the `"reason"` the python jsonschema spits out is universal, but it's useful for the python text and to provide some context.

pipermerriam: I think that it should include:

- validity flag
- JSON-pointers to invalid fields
   - Most cases can probably be captured with a single pointer.
   - There *might* be cases where we need multiple pointers.
   - Classification of the error.  This is probably something you need to make up.  This could be as simple as strings like `invalid-name`.  You can go further and define them using error codes like `N001 - Invalid Contract Type Name`, `N002 - Invalid source file path`.  (The `N` is just an arbitrary letter I choose to mimic the python linting error codes).
    - Maybe a human readable description.

I'd recommend starting simple and we can expand later since for a bit here we'll be the only consumers of these fixtures.

tests/test_fixtures.py
```diff
@@ -0,0 +1,33 @@
...
+from pathlib import Path
+import re
+import json
+from jsonschema import (
+    ValidationError as jsonValidationError,
+)
+
+# todo
+# validate manifest file formatting
+# use pytest.fixture to parametrize
```
pipermerriam, line 11: I'd look at using [`pytest_generate_tests`](https://docs.pytest.org/en/latest/parametrize.html#basic-pytest-generate-tests-example).  I believe it's the correct mechanism.  We use it in `py-evm/trinity` for generating tests from the `ethereum/tests` fixtures.

njgheorghita: ahh great, yeah that did the trick, thanks! this pr is ready for a final :eyes: if you have any other thoughts

PR #149, opened 2020-05-13
Add manifest formatting rules

njgheorghita:
Add fixtures testing manifest formatting rules

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -24,3 +24,21 @@ def validate_v3(schema):
     def _validate(*args, **kwargs):
         validate(args[0], schema, cls=validator_for(schema, Draft7Validator))
     return _validate
+
+@pytest.fixture
+def validate_manifest_formatting():
+    def _validate(*args, **kwargs):
+        raw_manifest = args[0]
+        try:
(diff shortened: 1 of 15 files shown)


tests/conftest.py
```diff
@@ -24,3 +24,21 @@ def validate_v3(schema):
...
+            manifest_dict = json.loads(raw_manifest)
+        except json.JSONDecodeError:
+            raise Exception("Invalid JSON")
+
+        if raw_manifest[-1:] == "\n":
+            raise Exception("Invalid trailing newline")
+
+        sorted_manifest = json.dumps(manifest_dict, sort_keys=True, separators=(",", ":"))
+        if raw_manifest != sorted_manifest:
+            raise Exception("Invalid manifest has unsorted keys, duplicate keys, or is not tightly packed.")
```
njgheorghita, line 42: It seems non-trivial to get more precise testing for these different cases, so I just lumped them together for now. But I can look into enhancing this test to explicitly cover each test case if it seems worthwhile. The "reason" in each fixture is more precise in what the particular cause of error is.

PR #150, opened 2020-05-15
Add strictly formatted v3 example manifests

njgheorghita:
Added properly formatted versions of the v3 examples.


njgheorghita:
merging since this is a minor update

PR #151, opened 2020-05-18
Update travis python version to 3.8

njgheorghita:
Failing test in ci due to outdated python version used by travis.


njgheorghita:
merging since this is a minor bugfix

PR #152, opened 2020-05-28
Update assets

njgheorghita:
Some changes addressed in this pr. 

- Updated example contracts to newest version of solidity (contract assets in their respective packages still need updating to reflect these changes, which is coming in a future pr)
- Minor bug in export markdown script
- Updated `use cases` docs to reflect v3 changes.
- Added a test case to detect invalid `manifest_version` fields in v3 manifests.
- Added an example (`simple-auction`) to show how the ethpm spec can be used for a compiled contract's metadata output.


njgheorghita:
closing this pr, to break it up into smaller prs

PR #154, opened 2020-06-23
Update example contracts to solidity 0.6.8

njgheorghita:
Updated the example solidity contracts to 0.6.8.

Updated the associated v3 manifests to reflect the updated contracts (ipfs hashes / deployments).


njgheorghita:
@pipermerriam Thanks! It seems like any hidden mistakes in here would be pretty impossible to find unless you want to spend a lot of time validating the ipfs hashes / bytecodes. I've been as careful as possible updating these, ran them through multiple validation functions (schema validation, and formatting validation), so they should be good to go!

PR #156, opened 2020-07-01
Fix invalid escrow deployments

njgheorghita:
Escrow manifest was invalid since it had two blockchain uris representing the same chain. Updated v3 strict and pretty manifests in this pr.


njgheorghita:
merging since this is just a minor bugfix pr

PR #158, opened 2020-07-07
Add example metadata compiler output for smart contract examples

njgheorghita:
Generated some reference examples for compiler teams who are planning to support ethpm natively in the [compiler metadata output](https://solidity.readthedocs.io/en/latest/metadata.html). Included in each example directory is now a `metadata/` folder that contains ethpm-compliant metadata output for the source contracts in each example.


njgheorghita:
@pipermerriam Yup, exactly!

PR #161, opened 2020-09-08
Rename ethpm uri fixture tests

njgheorghita:
Rename the ethpm uri fixture tests to reflect its respective EIP.


njgheorghita:
Merging since this is a minor filename change.

PR #171, opened 2021-10-05
fix: remove './' from sourceId's in examples

### LINKED ISSUE #170: Source ID mismatch in examples
In the examples, under each `ContractType` object, the `sourceId` field does not match anything under the `sources` field. Here are the `sourceId` fields for the `Escrow` example:
https://github.com/ethpm/ethpm-spec/blob/62a6ce2e7d6e10b28e02b0b42db56890f65486ce/examples/escrow/v3-pretty.json#L98
https://github.com/ethpm/ethpm-spec/blob/62a6ce2e7d6e10b28e02b0b42db56890f65486ce/examples/escrow/v3-pretty.json#L121

As you'll notice, these do not correspond with the keys of the `sources` object:
https://github.com/ethpm/ethpm-spec/blob/62a6ce2e7d6e10b28e02b0b42db56890f65486ce/examples/escrow/v3-pretty.json#L155
https://github.com/ethpm/ethpm-spec/blob/62a6ce2e7d6e10b28e02b0b42db56890f65486ce/examples/escrow/v3-pretty.json#L162

The solution here should be to rectify these two fields together by removing the base path (e.g. `./`), as the EIP specification does not mention an implicit assumption of the base path when referencing `sourceId` in the `sources` object.

fubuloubu:
fixes: #170


njgheorghita:
Thanks dog!

franciscoarturorivera371-cyber (commented):
Submit the change.

